PR #3, opened 2023-10-24
small fixes to formatting and typo

spiral-ladder:
the formatting change is slightly opinionated but (imo) it makes it easier to see the key-value options.


AbdelStark (approved):
it's indeed better

PR #23, opened 2023-10-25

### LINKED ISSUE #9: feat: add optional trace to CairoVM struct and add enable trace flag

itzlambda:
Fixes: #9 

I am new to zig so let me know if something is non idiomatic or there is a better way to do it.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -36,6 +36,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "tcoratger",
+      "name": "Thomas Coratger",
+      "avatar_url": "https://avatars.githubusercontent.com/u/60488569?v=4",
+      "profile": "https://github.com/tcoratger",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "contributorsPerLine": 7,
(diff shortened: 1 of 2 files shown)


src/vm/core.zig
```diff
@@ -102,6 +109,9 @@ pub const CairoVM = struct {
     // # Arguments
     /// - `instruction`: The instruction to run.
     pub fn runInstruction(self: *CairoVM, instruction: *const instructions.Instruction) !void {
+        if (self.trace != null) {
+            try self.trace.?.append(TraceEntry{ .pc = self.run_context.pc, .ap = self.run_context.ap, .fp = self.run_context.fp });
+        }
```
abdelhamidbakhta, line 114:
```suggestion
if(self.trace) |tracer| {
    try tracer.append(TraceEntry{ .pc = self.run_context.pc, .ap = self.run_context.ap, .fp = self.run_context.fp });
}
```

src/vm/core.zig
```diff
@@ -781,3 +797,84 @@ test "update fp dst felt" {
...
+    // ************************************************************
+    // Check that trace was initialized
+    const trace = vm.trace;
+
+    if (trace != null) {
+        return error.TraceShouldHaveBeenDisabled;
+    }
+}
+
+// TODO: Find a valid Instruction so we can call `runInstruction` to check if traces
```
abdelhamidbakhta, line 855: remove commented test and create an issue with the content of this test instead please

AbdelStark (changes requested):
left some comments

AbdelStark (commented):
I am wondering if we should just make the `config` struct of the cmd in a `config.zig` file and share it between the cli and the core vm. 
Instead of passing individual flags like we do in this PR.
I am worried that it will be harder to maintain if each time we add a new flag in the init function and a new field in the VM.
For example we see here that you had to change each individual test to add the new parameter.
If it was the struct and if the struct has a default function you would not have to change the tests most of the time, only when you need to change the config for an actual test.

lambda-0x:
reopened: #27

PR #27, opened 2023-10-26

### LINKED ISSUE #9: feat: add optional trace to CairoVM struct and add enable trace flag

itzlambda:
fix: #9

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -36,6 +36,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "tcoratger",
+      "name": "Thomas Coratger",
+      "avatar_url": "https://avatars.githubusercontent.com/u/60488569?v=4",
+      "profile": "https://github.com/tcoratger",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "contributorsPerLine": 7,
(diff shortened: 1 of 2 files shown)


AbdelStark (changes requested):
left comment

src/vm/core.zig
```diff
@@ -439,7 +471,11 @@ test "update pc regular with imm" {
     instruction.op_1_addr = instructions.Op1Src.Imm;
     const operands = OperandsResult.default();
     // Create a new VM instance.
-    var vm = try CairoVM.init(&allocator);
+    var config = Config{
```
abdelhamidbakhta, line 474: can you add a `default` function in `Config` and use it in tests ?

lambda-0x: right thats much better

bingcicle: As far as I recall you can also set defaults in the struct itself here:

https://github.com/keep-starknet-strange/cairo-zig/blob/3abd1bdfae155497ec9f46c8ff88f1b0b46c2183/src/vm/config.zig#L6-L8

Like so:

```zig
pub const Config = struct { 
     proof_mode: bool = false, 
     enable_trace: bool = false, 
```

and pass `.{}` everywhere:
```zig
var vm = try CairoVM.init(&allocator, .{});
```

lambda-0x: interesting!

lambda-0x:
haha lets try again: #29

PR #29, opened 2023-10-26

### LINKED ISSUE #9: feat: add optional trace to CairoVM struct and add enable trace flag

itzlambda:
fix: #9

### DIFF
--- a/src/vm/core.zig
+++ b/src/vm/core.zig
@@ -10,7 +10,7 @@ const relocatable = @import("memory/relocatable.zig");
 const instructions = @import("instructions.zig");
 const RunContext = @import("run_context.zig").RunContext;
 const CairoVMError = @import("error.zig").CairoVMError;
-const Config = @import("../cmd/cmd.zig").Config;
+const Config = @import("config.zig").Config;
 
 /// Represents the Cairo VM.
 pub const CairoVM = struct {
@@ -434,11 +434,7 @@ test "update pc regular no imm" {
     instruction.op_1_addr = instructions.Op1Src.AP;
     const operands = OperandsResult.default();
     // Create a new VM instance.
-    var config = Config{
-        .enable_trace = false,
-        .proof_mode = false,
-    };
-    var vm = try CairoVM.init(&allocator, config);
+    var vm = try CairoVM.init(&allocator, Config.default());
     defer vm.deinit();
 
     // ************************************************************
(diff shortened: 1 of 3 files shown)


src/vm/core.zig
```diff
@@ -37,17 +41,23 @@ pub const CairoVM = struct {
     /// - `CairoVM`: The created VM.
     /// # Errors
     /// - If a memory allocation fails.
-    pub fn init(allocator: *const Allocator) !CairoVM {
+    pub fn init(allocator: *const Allocator, config: Config) !CairoVM {
```
abdelhamidbakhta, line 44: add the argument in the doc of the function please

src/vm/core.zig
```diff
@@ -37,17 +41,23 @@ pub const CairoVM = struct {
...
         // Initialize the memory segment manager.
         const memory_segment_manager = try segments.MemorySegmentManager.init(allocator);
         // Initialize the run context.
         const run_context = try RunContext.init(allocator);
 
+        var trace: ?ArrayList(TraceEntry) = null;
+        if (config.enable_trace) {
+            trace = try ArrayList(TraceEntry).initCapacity(allocator.*, 4096);
```
abdelhamidbakhta, line 52: can you write a TODO comment with a link to the issue to make the initial capacity configurable please ?

src/vm/core.zig
```diff
@@ -913,3 +941,60 @@ test "update fp dst felt" {
...
+test "trace is enabled" {
+    // ************************************************************
+    // *                 SETUP TEST CONTEXT                       *
+    // ************************************************************
+    // Initialize an allocator.
+    var allocator = std.testing.allocator;
+
+    // Create a new VM instance.
+    var config = Config{ .proof_mode = false, .enable_trace = false };
+    config.enable_trace = true;
```
abdelhamidbakhta, line 954: why dont you put directly the true value when initiating ?

AbdelStark (changes requested):
left some comments

src/vm/core.zig
lambda-0x: oh initially i was using `Config.default()` there, then used regex to replace it :). will update.

src/vm/core.zig
lambda-0x: sorry missed that.

AbdelStark (approved):
lgtm

PR #33, opened 2023-10-26
feat(vm): make default `trace` `ArrayList` capacity configurable using `comptime`

### LINKED ISSUE #30: feat: make default `trace` ArrayList capacity configurable using comptime
https://github.com/keep-starknet-strange/cairo-zig/blob/ef11dff02cab138de0a6405901e87ad5a12f93fe/src/vm/core.zig#L54

currently `4096` value is hard coded, make this configurable at compile time using `comptime`

tcoratger:
Should close #30.

### DIFF
--- a/src/vm/core.zig
+++ b/src/vm/core.zig
@@ -42,7 +42,7 @@ pub const CairoVM = struct {
     /// - `CairoVM`: The created VM.
     /// # Errors
     /// - If a memory allocation fails.
-    pub fn init(allocator: *const Allocator, config: Config) !CairoVM {
+    pub fn init(comptime capacity: usize, allocator: *const Allocator, config: Config) !CairoVM {
         // Initialize the memory segment manager.
         const memory_segment_manager = try segments.MemorySegmentManager.init(allocator);
         // Initialize the run context.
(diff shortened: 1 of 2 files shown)


src/vm/core.zig
```diff
@@ -42,16 +42,15 @@ pub const CairoVM = struct {
     /// - `CairoVM`: The created VM.
     /// # Errors
     /// - If a memory allocation fails.
-    pub fn init(allocator: *const Allocator, config: Config) !CairoVM {
+    pub fn init(comptime capacity: usize, allocator: *const Allocator, config: Config) !CairoVM {
```
abdelhamidbakhta, line 45: please change to a more explicit name, it's more the `trace_entries_initial_capacity`
And I would put it after config in the orders of parameters.

AbdelStark (changes requested):
left small comments

src/vm/core.zig
tcoratger: @abdelhamidbakhta Sounds good, just pushed the change.

AbdelStark (approved):
lgtm

Issue #25, opened 2023-10-26
Optimize Tracing Implementation for Lower Runtime Overhead


AbdelStark:
### Summary
The current implementation proposed in #23  introduces a runtime overhead for each instruction to check if tracing is enabled or not. This issue discusses potential ways to optimize this.

### Problem
The existing method for tracing checks at runtime if the tracing feature is enabled, which incurs a small performance cost for every instruction in the VM.

#### 1. Use Inline Function with Comptime Check
- **What**: Create an inline function with a `comptime` check to see if tracing is activated.
- **Pros**: No runtime overhead as the check and branch won't be generated if tracing is disabled. Also, this is a good opportunity to explore Zig's `comptime` feature.
- **Cons**: The tracing flag is set at runtime via CLI. 
- **Mitigation**: Throw an error if CLI argument is true but the VM has not been compiled with tracing activated.

#### 2. Allocate Tracing Scratch Memory Upfront
- **What**: Pre-allocate some memory specifically for tracing.
- **Pros**: Amortizes the memory-growing cost. Reduces the overhead of array resizing (`ArrayList` resizes by doubling the memory when capacity is reached).
- **Cons**: Allocates memory that might go unused for small programs.
- **Mitigation**: Future versions could make this configurable for different scenarios.

### TL;DR
- Use an inline function with a `comptime` check for tracing to eliminate runtime overhead.
- Pre-allocate tracing scratch memory to amortize memory-growing costs.

### Questions
- What do you think about these proposed solutions? 
- Are there other trade-offs or mitigations to consider?

### Next Steps
- These optimizations can be made in two follow-up PRs.
- We can merge the current PR without these optimizations.

StringNick:
done through #514

PR #35, opened 2023-10-26
perf(vm): use dynamic dispatch to enable/disable tracing of instructions

nils-mathieu:
This PR introduces a dynamic dispatch mechanism to trace instructions within the VM.

This is related to #25.

# Performance Considerations

I'm not yet sure whether a function pointer call is that slower or faster than a regular if/else branch. The CPU might be able to optimze the latter better because of its knowledge of the content. With this PR, we might've just removed any optimization possibility entierly.

This should be benchmark to avoid regressions.

### DIFF
--- a/src/cmd/cmd.zig
+++ b/src/cmd/cmd.zig
@@ -11,6 +11,7 @@ const vm_core = @import("../vm/core.zig");
 const RunContext = @import("../vm/run_context.zig").RunContext;
 const relocatable = @import("../vm/memory/relocatable.zig");
 const Config = @import("../vm/config.zig").Config;
+const build_options = @import("../main.zig").build_options;
 
 // ************************************************************
 // *                 GLOBAL VARIABLES                         *
@@ -88,6 +89,11 @@ fn execute(_: []const []const u8) !void {
         "Running Cairo VM...\n",
         .{},
     );
+
+    if (build_options.disable_tracing and config.enable_trace) {
+        std.log.err("Tracing is disabled in this build.\n", .{});
+        return;
+    }
 
     // Create a new VM instance.
     var vm = try vm_core.CairoVM.init(&gpa_allocator, config);
--- a/src/cmd/cmd.zig
+++ b/src/cmd/cmd.zig
@@ -88,7 +88,7 @@ const UsageError = error{
     /// Occurs when the user requests a config that is not supported by the current build.
     ///
     /// For example because execution traces are globally disabled.
-    incompatibleBuildOptions,
+    IncompatibleBuildOptions,
 };
 
 // execute entrypoint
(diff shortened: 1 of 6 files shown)


nils-mathieu:
This remains is draft until I can figure out whether this is a regression or not :) This kind of optimization is really hard to predict.

src/cmd/cmd.zig
```diff
@@ -11,6 +11,7 @@ const vm_core = @import("../vm/core.zig");
 const RunContext = @import("../vm/run_context.zig").RunContext;
 const relocatable = @import("../vm/memory/relocatable.zig");
 const Config = @import("../vm/config.zig").Config;
+const build_options = @import("../main.zig").build_options;
```
abdelhamidbakhta, line 14: I would avoid using things from main because the VM can be used as a library and not always as an executable.

src/cmd/cmd.zig
```diff
@@ -89,6 +90,11 @@ fn execute(_: []const []const u8) !void {
         .{},
     );
 
+    if (build_options.disable_tracing and config.enable_trace) {
+        std.log.err("Tracing is disabled in this build.\n", .{});
+        return;
```
abdelhamidbakhta, line 95: better returning an error here no ?

src/vm/trace_context.zig
```diff
@@ -0,0 +1,132 @@
+const std = @import("std");
```
abdelhamidbakhta, line 1: please add the new `trace_context` file declaration in `lib.zig` for usage as lib as well as automatic fetching of unit tests, and generation of docs

AbdelStark (changes requested):
left some comments

src/cmd/cmd.zig
nils-mathieu: oooh I just saw that there was some constant options there and thought it would be the right place.

Where should I put this then? A separate "build_options.zig" file?

src/cmd/cmd.zig
nils-mathieu: Maybe?

I mean, the user probably wants to be properly warned about what's the problem. It's not like it's an obscure error. It's a user invalid usage I'd say.

src/cmd/cmd.zig
```diff
@@ -82,18 +83,30 @@ pub fn run() !void {
 // *                       CLI COMMANDS                       *
 // ************************************************************
 
+/// Errors that occur because the user misused the CLI.
+const UsageError = error{
+    /// Occurs when the user requests a config that is not supported by the current build.
+    ///
+    /// For example because execution traces are globally disabled.
+    incompatibleBuildOptions,
```
abdelhamidbakhta, line 91: please start with uppercase

AbdelStark (dismissed):
small comment

src/cmd/cmd.zig
nils-mathieu: Didn't even notice it lol. Way to used to writing everything in lowercase in C.

AbdelStark (approved):
lgtm

PR #40, opened 2023-10-26
Omit the struct type of a literal when the result is coerced

### LINKED ISSUE #36: Omit the struct type of a literal when the result is coerced
In Ziglang, following https://ziglang.org/documentation/0.11.0/#Anonymous-Struct-Literals,:

> Zig allows omitting the struct type of a literal. When the result is coerced, the struct literal will directly instantiate the result location, with no copy:

Thus, in functions where the output type is already mentioned in the function definition, there is no need to recall the name of the structure in the output. For example:

https://github.com/keep-starknet-strange/cairo-zig/blob/eda42da0d948f455a2f93aa83eee1193f1fbc9a4/src/math/fields/fields.zig#L126-L137

would become:

```ziglang
pub fn sub(
    self: Self,
    other: Self,
) Self {
    var ret: F.MontgomeryDomainFieldElement = undefined;
    F.sub(
        &ret,
        self.fe,
        other.fe,
    );
    return .{ .fe = ret };
}
```

This can be done to achieve a less verbose implementation and to follow the design of Ziglang for a less verbose implementation and to follow the design of the Ziglang std lib for example:

https://github.com/ziglang/zig/blob/a9e66ed730e17833a3f2701294c01f0b16d206e4/lib/std/io/c_writer.zig#L9-L11

tcoratger:
Should close #36.

### DIFF
--- a/src/math/fields/fields.zig
+++ b/src/math/fields/fields.zig
@@ -19,7 +19,7 @@ pub fn Field(
                 &bz,
                 [_]u8{0} ** BytesSize,
             );
-            break :val Self{ .fe = bz };
+            break :val .{ .fe = bz };
         };
 
         fe: F.MontgomeryDomainFieldElement,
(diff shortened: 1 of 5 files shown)


AbdelStark (approved):
Lgtm

PR #57, opened 2023-10-27
Use `Allocator` instead of `*const Allocator`.

### LINKED ISSUE #55: Use `Allocator` instead of `*const Allocator`.
Internally, `Allocator` is an opaque pointer and a VTable which are cheap to copy around. No need to use a reference.

For those knowing Rust, `Allocator` is like a `&dyn Allocator`. Meaning that `*const Allocator` is like a `&&dyn Allocator`.

In the codebase, I noticed serveral places where a `*const Allocator` was being passed instead of just passing it by value directly. It would be best to simply pass it as `Allocator`.

# Reference

The Zig standard library stores an `Allocator` directly in all of the standard types.

https://github.com/ziglang/zig/blob/8f48533691e93538846993f78f732272a03a600b/lib/std/array_list.zig#L46

tcoratger:
Should close #55.

### DIFF
--- a/src/cmd/cmd.zig
+++ b/src/cmd/cmd.zig
@@ -105,7 +105,7 @@ fn execute(_: []const []const u8) !void {
 
     // Create a new VM instance.
     var vm = try vm_core.CairoVM.init(
-        &gpa_allocator,
+        gpa_allocator,
         config,
     );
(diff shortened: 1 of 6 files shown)


AbdelStark (approved):
lgtm

PR #58, opened 2023-10-27
refactor: deprecate default fn from Relocatable

spiral-ladder:
We make use of Zig's default struct initialization to get rid of the default() fn.


AbdelStark (approved):
lgtm

PR #62, opened 2023-10-27
Li/feat/compute res

lana-shanghai:


src/vm/core.zig
```diff
@@ -196,6 +201,55 @@ pub const CairoVM = struct {
...
+    /// - `op_1`: The operand 1.
+    /// # Returns
+    /// - `res`: The result of the operation.
+    pub fn computeRes(self: *CairoVM, instruction: *const Instruction, op_0: relocatable.MaybeRelocatable, op_1: relocatable.MaybeRelocatable) error{ AddRelocToRelocForbidden, MulRelocForbidden, TypeMismatchNotFelt, TypeMismatchNotRelocatable, ValueTooLarge }!relocatable.MaybeRelocatable {
+        _ = self;
+
+        var res = switch (instruction.res_logic) {
+            // Simply return the op_1.
+            instructions.ResLogic.Op1 => op_1,
+            instructions.ResLogic.Add => {
```
abdelhamidbakhta, line 217: I would probably consider creating a function like this and simplify a bit to add more readibility:
```zig

Certainly, here's a simplified and commented version of the Zig code. I've combined some repeated steps and reorganized the conditions to make the code more concise and easier to read.

zig
Copy code
// Add two operands which can either be a "relocatable" or a "felt".
// The operation is allowed between:
// 1. A felt and another felt.
// 2. A felt and a relocatable.
// Adding two relocatables is forbidden.
pub fn addOperands(op_0: MaybeRelocatable, op_1: MaybeRelocatable) !Operand {
    // Both operands are relocatables, operation forbidden
    if (op_0.isRelocatable() and op_1.isRelocatable()) {
        return error.AddRelocToRelocForbidden;
    }

    // One of the operands is relocatable, the other is felt
    if (op_0.isRelocatable() or op_1.isRelocatable()) {
        // Determine which operand is relocatable and which one is felt
        const reloc_op = if (op_0.isRelocatable()) op_0 else op_1;
        const felt_op = if (op_0.isRelocatable()) op_1 else op_0;
        
        var reloc = try reloc_op.tryIntoRelocatable();
        var felt = try felt_op.tryIntoFelt();
        
        // Add the felt to the relocatable's offset
        try reloc.addFeltInPlace(felt);
        
        return relocatable.newFromRelocatable(reloc);
    }

    // Both operands are felts
    const op_0_felt = try op_0.tryIntoFelt();
    const op_1_felt = try op_1.tryIntoFelt();
    
    // Add the felts and return as a new felt wrapped in a relocatable
    return relocatable.fromFelt(op_0_felt.add(op_1_felt));
}
```
And call this function in the `instructions.ResLogic.Add` switch arm : `return try addOperands(op_0, op_1)`

AbdelStark (changes requested):
left some comments

src/vm/core.zig
```diff
@@ -991,3 +1050,121 @@ test "trace is disabled" {
...
+    try expectEqual(expected_value, actual_value);
+}
+
+test "compute res add works" {
+    // ************************************************************
+    // *                 SETUP TEST CONTEXT                       *
+    // ************************************************************
+    // Initialize an allocator.
+    var allocator = std.testing.allocator;
+    var instruction = Instruction{ .off_0 = 0, .off_1 = 1, .off_2 = 2, .dst_reg = instructions.Register.AP, .op_0_reg = instructions.Register.AP, .op_1_addr = instructions.Op1Src.AP, .res_logic = instructions.ResLogic.Add, .pc_update = instructions.PcUpdate.Regular, .ap_update = instructions.ApUpdate.Regular, .fp_update = instructions.FpUpdate.Regular, .opcode = instructions.Opcode.NOp };
```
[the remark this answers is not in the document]
lana-shanghai: LOL! I was wondering how to do that

src/vm/memory/segments.zig
```diff
@@ -119,19 +121,55 @@ test "memory segment manager" {
     const relocatable_address_1 = memory_segment_manager.addSegment();
 
     // Check that the memory segment manager has one segment.
-    try std.testing.expect(memory_segment_manager.memory.num_segments == 1);
+    try expect(memory_segment_manager.memory.num_segments == 1);
```
tcoratger, line 124: This is equivalent but there is a purpose built testing function for equality expectations, namely `std.testing.expectEqual`, see here for the full documentation: https://ziglang.org/documentation/0.11.0/#The-Testing-Namespace

src/vm/memory/segments.zig
```diff
@@ -119,19 +121,55 @@ test "memory segment manager" {
...
 
     // Check if the relocatable address is correct.
-    try std.testing.expect(relocatable_address_1.segment_index == 0);
-    try std.testing.expect(relocatable_address_1.offset == 0);
+    try expect(relocatable_address_1.segment_index == 0);
+    try expect(relocatable_address_1.offset == 0);
```
tcoratger, line 128: Here as we are testing the 2 elements of the structure which only has 2, would it not be better to make an expectation of equality on the complete structure. In this way, if we have to add fields to the structure one day, we will be forced to modify the test otherwise it will no longer pass and this completely secures the testing. To me it would look something like:

```ziglang
try std.testing.expectEqual(
  relocatable.Relocatable{
      .segment_index = 0,
      .offset = 0,
  },
relocatable_address_1,
)
```

src/vm/memory/segments.zig
```diff
@@ -119,19 +121,55 @@ test "memory segment manager" {
...
+    var memory_segment_manager = try MemorySegmentManager.init(allocator);
+    defer memory_segment_manager.deinit();
+
+    // ************************************************************
+    // *                      TEST BODY                           *
+    // ************************************************************
+    _ = memory_segment_manager.addSegment();
+    _ = memory_segment_manager.addSegment();
+
+    const address = relocatable.Relocatable.new(0, 0);
```
tcoratger, line 158: Would be good to add a comma at the end of two element lists for fmt:
```
const address = relocatable.Relocatable.new(0, 0,);
```

src/vm/core.zig
```diff
@@ -147,29 +149,32 @@ pub const CairoVM = struct {
...
-            .op_1 = relocatable.fromU64(0),
-            .dst_addr = relocatable.fromU64(0),
-            .op_0_addr = relocatable.fromU64(0),
-            .op_1_addr = relocatable.fromU64(0),
+        const res = try computeRes(instruction, op_0_op, op_1_op);
+
+        // Deduce the operands if they haven't been successfully retrieved from memory.
+        // TODO: Implement this.
+
+        return OperandsResult{
```
bingcicle, line 170: I believe this works?

```suggestion
        return .{
```

src/vm/core.zig
```diff
@@ -383,6 +388,100 @@ pub const CairoVM = struct {
...
+/// - `instruction`: The instruction to compute the operands for.
+/// - `op_0`: The operand 0.
+/// - `op_1`: The operand 1.
+/// # Returns
+/// - `res`: The result of the operation.
+pub fn computeRes(
+    instruction: *const Instruction,
+    op_0: relocatable.MaybeRelocatable,
+    op_1: relocatable.MaybeRelocatable,
+) CairoVMError!relocatable.MaybeRelocatable {
```
bingcicle, line 402: To avoid these long types we can import `MaybeRelocatable` above like so:

```zig
const MaybeRelocatable = relocatable.MaybeRelocatable;
```
Then this can be:
```suggestion
    op_0: MaybeRelocatable,
    op_1: MaybeRelocatable,
) CairoVMError!MaybeRelocatable {
```

Ditto for the others

src/vm/core.zig
```diff
@@ -383,6 +388,100 @@ pub const CairoVM = struct {
...
+    // One of the operands is a relocatable, operation forbidden
+    if (op_0.isRelocatable() and op_1.isRelocatable()) {
+        return CairoVMError.MulRelocForbidden;
+    }
+
+    // One of the operands is relocatable, the other is felt
+    if (op_0.isRelocatable() or op_1.isRelocatable()) {
+        return CairoVMError.MulRelocForbidden;
+    }
```
bingcicle, line 478:
```suggestion
    // If one of the operands is relocatable, operation forbidden
    if (op_0.isRelocatable() or op_1.isRelocatable()) {
        return CairoVMError.MulRelocForbidden;
    }
```

If both the AND and OR cases are forbidden, it seems like checking OR is enough?

lana-shanghai: agree

src/vm/core.zig
lana-shanghai: It does! Magic

src/vm/core.zig
```diff
@@ -383,19 +389,110 @@ pub const CairoVM = struct {
...
+/// - `op_1`: The operand 1.
+/// # Returns
+/// - `res`: The result of the operation.
+pub fn computeRes(
+    instruction: *const Instruction,
+    op_0: MaybeRelocatable,
+    op_1: MaybeRelocatable,
+) CairoVMError!MaybeRelocatable {
+    var res = switch (instruction.res_logic) {
+        instructions.ResLogic.Op1 => op_1,
```
tcoratger, line 405: Here I think that `.ResLogic.Op1 => op_1` works

src/vm/core.zig
```diff
@@ -383,19 +389,110 @@ pub const CairoVM = struct {
...
-    op_0: relocatable.MaybeRelocatable,
-    op_1: relocatable.MaybeRelocatable,
-    dst_addr: relocatable.MaybeRelocatable,
-    op_0_addr: relocatable.MaybeRelocatable,
-    op_1_addr: relocatable.MaybeRelocatable,
+    dst: MaybeRelocatable,
+    res: ?MaybeRelocatable,
+    op_0: MaybeRelocatable,
+    op_1: MaybeRelocatable,
+    dst_addr: relocatable.Relocatable,
```
tcoratger, line 493: Here as you import a `MaybeRelocatable` constant type at the beginning of the file, I think we can also import `Relocatable` as `relocatable.Relocatable`

lana-shanghai: done

src/vm/builtins/bitwise/bitwise.zig
```diff
@@ -140,7 +140,10 @@ test "deduce when address points to felt greater than BITWISE_TOTAL_N_BITS" {
     // when
     var address = Relocatable.new(0, 3);
 
-    try mem.set(Relocatable.new(0, 0), fromFelt(Felt252.fromInteger(number)));
+    try mem.set(Relocatable.new(
+        0,
+        0,
+    ), fromFelt(Felt252.fromInteger(number)));
```
abdelhamidbakhta, line 146: use `fromU256` directly

src/vm/core.zig
```diff
@@ -147,29 +151,32 @@ pub const CairoVM = struct {
         // Compute the destination address and get value from the memory.
         const dst_addr = try self.run_context.compute_dst_addr(instruction);
         const dst = try self.segments.memory.get(dst_addr);
-        _ = dst;
 
         // Compute the OP 0 address and get value from the memory.
         const op_0_addr = try self.run_context.compute_op_0_addr(instruction);
         // Here we use `catch null` because we want op_0_op to be optional since it's not always used.
-        const op_0_op = self.segments.memory.get(op_0_addr) catch null;
+        const op_0_op = try self.segments.memory.get(op_0_addr);
```
abdelhamidbakhta, line 158: are you sure of this ? it seems it should be allow to be null, with the `try` I think it breaks the implementation for some cases. worth checking

AbdelStark (changes requested):
left some comments

src/vm/core.zig
lana-shanghai: I would keep it as a TODO because we still do not have the deduce op0 and compute op0 operands

abdelhamidbakhta: sounds good

AbdelStark (approved):

src/vm/core.zig
bingcicle: On hindsight we could even remove the comment if you wanted. The if block as well as the named error is clear enough such that the comment seems redundant, but this is just a nitpick at this point, so ignore me please :)

Issue #46, opened 2023-10-27
feat: implement deduce_op_1


lana-shanghai:
### Description

Currently the `runInstruction` function doesn't work because we are missing some of the core logic of the state transition functions. In `runInstruction` we call `computeOperands` which calls `computeOp1Deductions` which tries to deduce `op_1` memory cell from builtins, if it fails (right now it returns `null` by default), it tries to deduce the operand from optional `dst` and `op_0`. 

Python [implementation](https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/cairo/lang/vm/vm_core.py#L214) 

### Acceptance Criteria

Add tests similar to the Rust [implementation](https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/vm/vm_core.rs#L2084).

nils-mathieu:
I'd love to work on that!

PR #69, opened 2023-10-27
feat: Implementation of `deduceOp1`

nils-mathieu:
This PR solves #46.

# Context

Re-implementation of the [Python implementation of the same function](https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/cairo/lang/vm/vm_core.py#L214).

### DIFF
--- a/src/vm/core.zig
+++ b/src/vm/core.zig
@@ -358,49 +358,6 @@ pub const CairoVM = struct {
         }
     }
 
-    /// Attempts to deduce `op1` and `res` for an instruction, given `dst` and `op0`.
-    ///
-    /// # Arguments
-    /// - `inst`: The instruction to deduce `op1` and `res` for.
-    /// - `dst`: The destination of the instruction.
-    /// - `op0`: The first operand of the instruction.
-    ///
-    /// # Returns
-    /// - `Tuple`: A tuple containing the deduced `op1` and `res`.
-    fn deduce_op1(
-        self: *const CairoVM,
-        inst: *const instructions.Instruction,
-        dst: ?*const relocatable.MaybeRelocatable,
-        op0: ?*const relocatable.MaybeRelocatable,
-    ) !std.meta.Tuple(&[_]type{ ?relocatable.MaybeRelocatable, ?relocatable.MaybeRelocatable }) {
-        _ = self;
-
-        if (inst.opcode != .AssertEq) {
-            return .{ null, null };
-        }
-
-        switch (inst.res_logic) {
-            .Op1 => if (dst) |dst_val| {
-                return .{ dst_val.*, dst_val.* };
-            },
-            .Add => if (dst != null and op0 != null) {
-                return .{ try dst.?.sub(op0.?.*), dst.?.* };
-            },
-            .Mul => {
-                if (dst == null or op0 == null)
-                    return .{ null, null };
-                if (dst.?.isRelocatable() or op0.?.isRelocatable())
-                    return .{ null, null };
-                if (op0.?.felt.isZero())
-                    return .{ null, null };
-                return .{ relocatable.fromFelt(try dst.?.felt.div(op0.?.felt)), dst.?.* };
-            },
-            else => {},
-        }
-
-        return .{ null, null };
-    }
-
     // ************************************************************
     // *                    ACCESSORS                             *
     // ************************************************************
@@ -523,6 +480,46 @@ pub fn mulOperands(
 
     // Multiply the felts and return as a new felt wrapped in a relocatable
     return relocatable.fromFelt(op_0_felt.mul(op_1_felt));
+}
+
+/// Attempts to deduce `op1` and `res` for an instruction, given `dst` and `op0`.
+///
+/// # Arguments
+/// - `inst`: The instruction to deduce `op1` and `res` for.
+/// - `dst`: The destination of the instruction.
+/// - `op0`: The first operand of the instruction.
+///
+/// # Returns
+/// - `Tuple`: A tuple containing the deduced `op1` and `res`.
+pub fn deduce_op1(
+    inst: *const instructions.Instruction,
+    dst: ?*const relocatable.MaybeRelocatable,
+    op0: ?*const relocatable.MaybeRelocatable,
+) !std.meta.Tuple(&[_]type{ ?relocatable.MaybeRelocatable, ?relocatable.MaybeRelocatable }) {
+    if (inst.opcode != .AssertEq) {
+        return .{ null, null };
+    }
+
+    switch (inst.res_logic) {
+        .Op1 => if (dst) |dst_val| {
+            return .{ dst_val.*, dst_val.* };
+        },
+        .Add => if (dst != null and op0 != null) {
+            return .{ try dst.?.sub(op0.?.*), dst.?.* };
+        },
+        .Mul => {
+            if (dst == null or op0 == null)
+                return .{ null, null };
+            if (dst.?.isRelocatable() or op0.?.isRelocatable())
+                return .{ null, null };
+            if (op0.?.felt.isZero())
+                return .{ null, null };
+            return .{ relocatable.fromFelt(try dst.?.felt.div(op0.?.felt)), dst.?.* };
+        },
+        else => {},
+    }
+
+    return .{ null, null };
 }
 
 // *****************************************************************************
@@ -1179,9 +1176,7 @@ test "deduce_op1 when opcode == .AssertEq, input is felt" {
     // ************************************************************
     // *                 SETUP TEST CONTEXT                       *
     // ************************************************************
-    var allocator = std.testing.allocator;
-    var vm = try CairoVM.init(allocator, .{});
-    defer vm.deinit();
+    // Nothing.
 
     // ************************************************************
     // *                      TEST BODY                           *
@@ -1202,7 +1197,7 @@ test "deduce_op1 when opcode == .AssertEq, input is felt" {
 
     const dst = relocatable.fromU64(3);
     const op0 = relocatable.fromU64(2);
-    const op1_and_result = try vm.deduce_op1(&instruction, &dst, &op0);
+    const op1_and_result = try deduce_op1(&instruction, &dst, &op0);
     const op1 = op1_and_result[0];
     const res = op1_and_result[1];
 
@@ -1217,9 +1212,7 @@ test "deduce_op1 when opcode == .AssertEq, res_logic == .Mul, non-zero op0" {
     // ************************************************************
     // *                 SETUP TEST CONTEXT                       *
     // ************************************************************
-    var allocator = std.testing.allocator;
-    var vm = try CairoVM.init(allocator, .{});
-    defer vm.deinit();
+    // Nothing.
 
     // ************************************************************
     // *                      TEST BODY                           *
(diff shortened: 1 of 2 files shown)


src/vm/core.zig
```diff
@@ -358,6 +358,56 @@ pub const CairoVM = struct {
...
+    /// - `inst`: The instruction to deduce `op1` and `res` for.
+    ///
+    /// - `dst`: The destination of the instruction.
+    ///
+    /// - `op0`: The first operand of the instruction.
+    ///
+    /// # Returns
+    ///
+    /// - `Tuple`: A tuple containing the deduced `op1` and `res`.
+    fn deduce_op1(self: *const CairoVM, inst: *const instructions.Instruction, dst: ?*const relocatable.MaybeRelocatable, op0: ?*const relocatable.MaybeRelocatable) !std.meta.Tuple(&[_]type{ ?relocatable.MaybeRelocatable, ?relocatable.MaybeRelocatable }) {
```
tcoratger, line 374: Would be good to just add a comma at the end of this list for fmt :)

src/vm/core.zig
```diff
@@ -358,6 +358,56 @@ pub const CairoVM = struct {
...
+        _ = self;
```
tcoratger, line 375: Would be good to have that as a function and not as a method to avoid this, what do you think?

src/vm/core.zig
nils-mathieu: Yep, just realized zig formatted it like this. 250 columns ez

src/vm/core.zig
nils-mathieu: I followed what the python implementation was doing, but I planned to ask about this as well once I'm settled with the actual implementation.

But yeah, this should probably not be a member.

Do I just move it out of the struct definition?

nils-mathieu: Actually, I just saw other funcitons of the same kind outside of the struct. That makes much sense x)

src/vm/core.zig
```diff
@@ -482,6 +482,46 @@ pub fn mulOperands(
...
+/// Attempts to deduce `op1` and `res` for an instruction, given `dst` and `op0`.
+///
+/// # Arguments
+/// - `inst`: The instruction to deduce `op1` and `res` for.
+/// - `dst`: The destination of the instruction.
+/// - `op0`: The first operand of the instruction.
+///
+/// # Returns
+/// - `Tuple`: A tuple containing the deduced `op1` and `res`.
+pub fn deduce_op1(
```
bingcicle, line 494: camelCase for fns
```suggestion
pub fn deduceOp1(
```

nils-mathieu: Rust haunting me

bingcicle: we can keep our fingers crossed for that proposal for snake_case functions to be accepted and implemented, because I do agree that snake_case fns are probably better :)

src/vm/core.zig
```diff
@@ -1096,6 +1136,254 @@ test "trace is disabled" {
...
+    // *                 SETUP TEST CONTEXT                       *
+    // ************************************************************
+    var allocator = std.testing.allocator;
+    var vm = try CairoVM.init(allocator, .{});
+    defer vm.deinit();
+
+    // ************************************************************
+    // *                      TEST BODY                           *
+    // ************************************************************
+    const instruction = instructions.Instruction{
```
bingcicle, line 1150: We are repeating this instruction for every test - we can also bring this into the global scope and change its values for each test. Something like:

```zig
var instruction.= Instruction { ... }

test "deduceOp1 when opcode == .Call" {
    instruction.opcode = .Call;
 
    // proceed with test
}
```

src/vm/core.zig
```diff
@@ -482,6 +482,46 @@ pub fn mulOperands(
...
+        .Mul => {
+            if (dst == null or op0 == null)
+                return .{ null, null };
+            if (dst.?.isRelocatable() or op0.?.isRelocatable())
+                return .{ null, null };
+            if (op0.?.felt.isZero())
+                return .{ null, null };
+            return .{ relocatable.fromFelt(try dst.?.felt.div(op0.?.felt)), dst.?.* };
+        },
+        else => {},
```
bingcicle, line 519: I think this could be simplified further to return nulls at one place, but i don't have a better idea at this moment

src/vm/core.zig
```diff
@@ -1096,6 +1136,254 @@ test "trace is disabled" {
...
+
+    const op1_and_result = try deduceOp1(&instruction, null, null);
+    const op1 = op1_and_result[0];
+    const res = op1_and_result[1];
+
+    // ************************************************************
+    // *                      TEST CHECKS                         *
+    // ************************************************************
+    try expectEqual(op1, null);
+    try expectEqual(res, null);
```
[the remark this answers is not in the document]
nils-mathieu: Oh! That's great! Didn't know about this.

src/vm/core.zig
```diff
@@ -1096,6 +1136,254 @@ test "trace is disabled" {
...
+        .off_2 = 3,
+        .dst_reg = .FP,
+        .op_0_reg = .AP,
+        .op_1_addr = .AP,
+        .res_logic = .Add,
+        .pc_update = .Jump,
+        .ap_update = .Regular,
+        .fp_update = .Regular,
+        .opcode = .Call,
+    };
```
bingcicle, line 1162: We are creating a new `Instruction` with all its key-value pairs for each test. We can also bring this up one level and change the values we are interested in for each test.

Something like:

```zig
var test_instruction = Instruction { ... };

test "deduceOp1 ..." {
    test_instruction.opcode = .Call;

    // rest of test
}
```

src/vm/core.zig
nils-mathieu: It'a a bit better than what I had before, but yeah. Still not ideal.

The easiest is to do something like this:

```zig
if (dst != null and op0 != null and !dst.?.isRelocatable() and !op0.?.isRelocatable() and !op0.?.felt.isZero())
    return .{ relocatable.fromFelt(try dst.?.felt.div(op0.?.felt)), dst.?.* };
```

But the auto formatter keeps putting this on one line, making it hard to read and hard to reason about properly.

src/vm/core.zig
nils-mathieu: duplicate?

but yeah agreed, I'll fix that

nils-mathieu:
It's not clear while the tests are not compiling in CI. Looking at the trace, I noticed it's using zig 0.12.0, while I'm using zig 0.11.0.

I'll update but it seems to be something else.

src/vm/core.zig
bingcicle: That should work if you manually introduce newlines after the 'and's, and I think you might have to put a comma after `dst.?.*`

nils-mathieu: oooh the `and` has to go at the end of the line!

Should I change it?

src/vm/core.zig
```diff
@@ -1096,6 +1138,203 @@ test "trace is disabled" {
     }
 }
 
+// This instruction is used in the functions that test the `deduceOp1` function. Only the
+// `opcode` and `res_logic` fields are usually changed.
+const deduceOp1TestInstr = instructions.Instruction{
```
bingcicle, line 1143:
```suggestion
var deduce_op1_test_instr = instructions.Instruction{
```

src/vm/core.zig
```diff
@@ -1096,6 +1138,203 @@ test "trace is disabled" {
...
+    // ************************************************************
+    // *                 SETUP TEST CONTEXT                       *
+    // ************************************************************
+    // Nothing.
+
+    // ************************************************************
+    // *                      TEST BODY                           *
+    // ************************************************************
+    var instr = deduceOp1TestInstr;
+    instr.opcode = .Call;
```
bingcicle, line 1167: If we declare the original instr as `var`, we can do this:

```suggestion
    deduce_op1_test_instr.opcode = .Call;
```

src/vm/memory/relocatable.zig
```diff
@@ -224,6 +252,18 @@ pub const MaybeRelocatable = union(enum) {
...
+    /// Returns whether the MaybeRelocatable is a felt or not.
+    ///
+    /// # Returns
+    ///
+    /// `true` if the MaybeRelocatable is a felt, `false` otherwise.
+    pub fn isFelt(self: MaybeRelocatable) bool {
+        return switch (self) {
+            .relocatable => false,
+            .felt => true,
+        };
```
bingcicle, line 265:
```suggestion
        return std.meta.activeTag(self) == .felt
```

spiral-ladder (commented):
just a few more nits

src/vm/core.zig
```diff
@@ -1096,6 +1138,203 @@ test "trace is disabled" {
...
+    const tuple = try deduceOp1(&instr, null, null);
+    const op1 = tuple[0];
+    const res = tuple[1];
+
+    // ************************************************************
+    // *                      TEST CHECKS                         *
+    // ************************************************************
+
+    try expectEqual(op1, null);
+    try expectEqual(res, null);
```
tcoratger, line 1178: Would be `try expectEqual(null, res);`, related #78.

nils-mathieu: Oh! Good catch.

src/vm/memory/relocatable.zig
nils-mathieu: Oh didn't know about `.activeTag`. That's great.

I'll change the `.isRelocatable` function as well.

src/vm/core.zig
nils-mathieu: Yeah but that creates a race condition if tests execute in parallel

src/vm/memory/relocatable.zig
```diff
@@ -172,6 +183,23 @@ pub const MaybeRelocatable = union(enum) {
         };
     }
 
+    /// Subtracts a `MaybeRelocatable` from this one and returns the new value.
+    ///
+    /// Only values of the same type may be subtracted. Specifically, attempting to
+    /// subtract a `.felt` with a `.relocatable` will result in an error.
+    pub fn sub(self: MaybeRelocatable, other: MaybeRelocatable) !MaybeRelocatable {
```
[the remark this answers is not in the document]
nils-mathieu: Pph! I didn't see that.

+ how did I miss this comment? I swear it did not show up

src/vm/memory/relocatable.zig
```diff
@@ -222,11 +233,17 @@ pub const MaybeRelocatable = union(enum) {
...
+    pub fn isRelocatable(self: MaybeRelocatable) bool {
+        return std.meta.activeTag(self) == .relocatable;
+    }
+
+    /// Returns whether the MaybeRelocatable is a felt or not.
+    ///
+    /// # Returns
+    ///
+    /// `true` if the MaybeRelocatable is a felt, `false` otherwise.
+    pub fn isFelt(self: MaybeRelocatable) bool {
```
lana-shanghai, line 245: good point - do you mind also replacing `!x.isRelocatable()` with `x.isFelt()` where relevant?

nils-mathieu: sure!

lana-shanghai (approved):
LGTM

PR #76, opened 2023-10-27
Use `Self` type inside structs/enum/unions

### LINKED ISSUE #71: Use `Self` type inside structs/enum/unions
At the moment, several `struct`/`enum`/`union` contain the name of the `struct`/`enum`/`union` duplicated in the different implementations of the object concerned. To follow Zig standards, it is best to declare `const Self = @This();` at the beginning of `struct`/`enum`/`union` and use it in all implementations inside the `struct`/`enum`/`union`. This allows for example to change the name of a structure with much fewer changes to make and it is also easier to read.

For example,
https://github.com/keep-starknet-strange/cairo-zig/blob/53f6a5b1053a3e122e4c86d0423ac3050f52e68c/src/vm/core.zig#L126-L129

would be

```
pub fn runInstruction(
        self: *Self,
        instruction: *const instructions.Instruction,
    ) !void {
```

with `const Self = @This();` declared at the beginning of the `CairoVM` struct.

tcoratger:
Should close #71.

### DIFF
--- a/src/vm/core.zig
+++ b/src/vm/core.zig
@@ -20,6 +20,7 @@ const Instruction = @import("instructions.zig").Instruction;
 
 /// Represents the Cairo VM.
 pub const CairoVM = struct {
+    const Self = @This();
 
     // ************************************************************
     // *                        FIELDS                            *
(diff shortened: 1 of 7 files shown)


AbdelStark (approved):
Lgtm

PR #83, opened 2023-10-28
fix: memory leaks in case of error during initialization

nils-mathieu:
Functions like this one (initialization of `CairoVM`)

```zig
pub fn init(
    allocator: Allocator,
    config: Config,
) !Self {
    // Initialize the memory segment manager.
    const memory_segment_manager = try segments.MemorySegmentManager.init(allocator);
    // Initialize the run context.
    const run_context = try RunContext.init(allocator);
    // Initialize the trace context.
    const trace_context = try TraceContext.init(allocator, config.enable_trace);

    return Self{
        .allocator = allocator,
        .run_context = run_context,
        .builtin_runners = ArrayList(BuiltinRunner).init(allocator),
        .segments = memory_segment_manager,
        .is_run_finished = false,
        .trace_context = trace_context,
    };
}
```

... can leak memory if one of the allocations fail (specifically, the first two).

In order to properly cleanup when the one of those allocations fail, we need to add some `errdefer`s.

```zig
pub fn init(
    allocator: Allocator,
    config: Config,
) !Self {
    // Initialize the memory segment manager.
    const memory_segment_manager = try segments.MemorySegmentManager.init(allocator);
    errdefer memory_segment_manager.deinit();
    // Initialize the run context.
    const run_context = try RunContext.init(allocator);
    errdefer run_context.deinit();
    // Initialize the trace context.
    const trace_context = try TraceContext.init(allocator, config.enable_trace);
    errdefer trace_context.deinit(); // technically not needed, but probably cleaner with it

    return Self{
        .allocator = allocator,
        .run_context = run_context,
        .builtin_runners = ArrayList(BuiltinRunner).init(allocator),
        .segments = memory_segment_manager,
        .is_run_finished = false,
        .trace_context = trace_context,
    };
}
```

This PR adds those `errdefer`s.

### DIFF
--- a/src/vm/core.zig
+++ b/src/vm/core.zig
@@ -67,11 +67,14 @@ pub const CairoVM = struct {
         // Initialize the trace context.
         const trace_context = try TraceContext.init(allocator, config.enable_trace);
         errdefer trace_context.deinit();
+        // Initialize the built-in runners.
+        const builtin_runners = ArrayList(BuiltinRunner).init(allocator);
+        errdefer builtin_runners.deinit();
 
         return Self{
             .allocator = allocator,
             .run_context = run_context,
-            .builtin_runners = ArrayList(BuiltinRunner).init(allocator),
+            .builtin_runners = builtin_runners,
             .segments = memory_segment_manager,
             .is_run_finished = false,
             .trace_context = trace_context,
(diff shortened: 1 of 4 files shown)


AbdelStark (approved):
lgtm

PR #84, opened 2023-10-28
feat(tests): fix argument order in `expectEqual` tests

### LINKED ISSUE #78: Reorder arguments in some `expectEqual` unit tests
I discovered a few days ago that, in reverse to the Rust syntax for example, the `expectEqual` function used in unit tests is in the form: `pub fn expectEqual(expected: anytype, actual: @TypeOf (expected))`. Which means the expected element is first and the current element to be tested is second.

This is not serious in itself but it makes it easier to debug console error messages. I saw that there were quite a few unit tests currently in the code where this order was not respected. It would be good to change that, I think, so everything is clear for the future.

For example,

https://github.com/keep-starknet-strange/cairo-zig/blob/5f70441423f74b3238dfbc0ac8db8b31f769c4b8/src/vm/builtins/bitwise/bitwise.zig#L191

would become

```
try expectEqual(expected, result);
```

tcoratger:
Should close #78.

### DIFF
--- a/src/math/fields/starknet.zig
+++ b/src/math/fields/starknet.zig
@@ -10,80 +10,89 @@ pub const Felt252 = fields.Field(
     0x800000000000011000000000000000000000000000000000000000000000001,
 );
 
+const expect = std.testing.expect;
+const expectEqual = std.testing.expectEqual;
+
 test "Felt252 fromInteger" {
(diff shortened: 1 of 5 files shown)


AbdelStark (approved):
lgtm, just need a conflict solving

PR #86, opened 2023-10-29
refac: various refactoring and safer behaviour for casting between integer types

tcoratger:
This PR contains:
- Several small refactorings for less verbose code in certain small places.
- Safer behavior for integer casting by systematically using `@intCast`

Please let me know if there are any items that are not suitable or need improvement.


AbdelStark (approved):
lgtm

PR #87, opened 2023-10-30
feat: initial implementation of `deduceOp0`

### LINKED ISSUE #45: feat: implement deduce_op_0
### Description

Currently the `runInstruction` function doesn't work because we are missing some of the core logic of the state transition functions. In `runInstruction` we call `computeOperands` which calls `computeOp0Deductions` which tries to deduce `op_0` memory cell from builtins, if it fails (right now it returns `null` by default), it tries to deduce the operand from optional `dst` and `op_1`. 

Python [implementation](https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/cairo/lang/vm/vm_core.py#L189) 

### Acceptance Criteria

Add tests similar to the Rust [implementation](https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/vm/vm_core.rs#L1877).

itzlambda:
fixes: #45 

TODO:
- [ ] update comment documentation

### DIFF
--- a/src/vm/core.zig
+++ b/src/vm/core.zig
@@ -219,38 +219,43 @@ pub const CairoVM = struct {
         _ = op_o;
     }
 
-    /// Attempts to deduce `op0` and `res` for an instruction, given `dst` and `op0`.
+    /// Attempts to deduce `op0` and `res` for an instruction, given `dst` and `op1`.
     ///
     /// # Arguments
     /// - `inst`: The instruction to deduce `op0` and `res` for.
     /// - `dst`: The destination of the instruction.
-    /// - `op0`: The first operand of the instruction.
+    /// - `op1`: The first operand of the instruction.
     ///
     /// # Returns
     /// - `Tuple`: A tuple containing the deduced `op0` and `res`.
     pub fn deduceOp0(
         self: *Self,
         inst: *const instructions.Instruction,
-        dst: ?*const relocatable.MaybeRelocatable,
-        op1: ?*const relocatable.MaybeRelocatable,
-    ) !std.meta.Tuple(&[_]type{ ?relocatable.MaybeRelocatable, ?relocatable.MaybeRelocatable }) {
-        if (inst.opcode == .Call) {
-            var result: ?MaybeRelocatable = relocatable.newFromRelocatable(try self.run_context.pc.addUint(inst.size()));
-            return .{ result, null };
-        } else if (inst.opcode == .AssertEq) {
-            if ((inst.res_logic == .Add) and (dst != null) and (op1 != null)) {
-                return .{ try subOperands(dst.?.*, op1.?.*), dst.?.* };
-            } else if (dst != null and op1 != null and
-                dst.?.isFelt() and op1.?.isFelt() and
-                !op1.?.felt.isZero())
-            {
-                return .{
-                    relocatable.fromFelt(try dst.?.felt.div(op1.?.felt)),
-                    dst.?.*,
-                };
-            }
+        dst: ?*const MaybeRelocatable,
+        op1: ?*const MaybeRelocatable,
+    ) !std.meta.Tuple(&.{ ?relocatable.MaybeRelocatable, ?relocatable.MaybeRelocatable }) {
+        switch (inst.opcode) {
+            .Call => {
+                return .{ relocatable.newFromRelocatable(try self.run_context.pc.addUint(inst.size())), null };
+            },
+            .AssertEq => {
+                const dst_val = dst orelse return .{ null, null };
+                const op1_val = op1 orelse return .{ null, null };
+                if ((inst.res_logic == .Add)) {
+                    return .{ try subOperands(dst_val.*, op1_val.*), dst_val.* };
+                } else if (dst_val.isFelt() and op1_val.isFelt() and
+                    !op1_val.felt.isZero())
+                {
+                    return .{
+                        relocatable.fromFelt(try dst_val.felt.div(op1_val.felt)),
+                        dst_val.*,
+                    };
+                }
+            },
+            else => {
+                return .{ null, null };
+            },
         }
-
         return .{ null, null };
     }
 
@@ -570,8 +575,8 @@ pub fn subOperands(self: MaybeRelocatable, other: MaybeRelocatable) !MaybeReloca
 /// - `Tuple`: A tuple containing the deduced `op1` and `res`.
 pub fn deduceOp1(
     inst: *const instructions.Instruction,
-    dst: ?*const relocatable.MaybeRelocatable,
-    op0: ?*const relocatable.MaybeRelocatable,
+    dst: ?*const MaybeRelocatable,
+    op0: ?*const MaybeRelocatable,
 ) !std.meta.Tuple(&[_]type{ ?relocatable.MaybeRelocatable, ?relocatable.MaybeRelocatable }) {
     if (inst.opcode != .AssertEq) {
         return .{ null, null };
--- a/src/vm/core.zig
+++ b/src/vm/core.zig
@@ -219,6 +219,41 @@ pub const CairoVM = struct {
         _ = op_o;
     }
 
+    /// Attempts to deduce `op0` and `res` for an instruction, given `dst` and `op0`.
+    ///
+    /// # Arguments
+    /// - `inst`: The instruction to deduce `op0` and `res` for.
+    /// - `dst`: The destination of the instruction.
+    /// - `op0`: The first operand of the instruction.
+    ///
+    /// # Returns
+    /// - `Tuple`: A tuple containing the deduced `op0` and `res`.
+    pub fn deduceOp0(
+        self: *Self,
+        inst: *const instructions.Instruction,
+        dst: ?*const relocatable.MaybeRelocatable,
+        op1: ?*const relocatable.MaybeRelocatable,
+    ) !std.meta.Tuple(&[_]type{ ?relocatable.MaybeRelocatable, ?relocatable.MaybeRelocatable }) {
+        if (inst.opcode == .Call) {
+            var result: ?MaybeRelocatable = relocatable.newFromRelocatable(try self.run_context.pc.addUint(inst.size()));
+            return .{ result, null };
+        } else if (inst.opcode == .AssertEq) {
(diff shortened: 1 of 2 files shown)


src/vm/core.zig
```diff
@@ -213,6 +213,41 @@ pub const CairoVM = struct {
...
+    /// - `inst`: The instruction to deduce `op0` and `res` for.
+    /// - `dst`: The destination of the instruction.
+    /// - `op0`: The first operand of the instruction.
+    ///
+    /// # Returns
+    /// - `Tuple`: A tuple containing the deduced `op0` and `res`.
+    pub fn deduceOp0(
+        self: *Self,
+        inst: *const instructions.Instruction,
+        dst: ?*const relocatable.MaybeRelocatable,
```
tcoratger, line 228: Here I think that we can use directly `MaybeRelocatable` (instead of `relocatable.MaybeRelocatable`) which is declared as a type at the beginning of the file.

src/vm/core.zig
```diff
@@ -213,6 +213,41 @@ pub const CairoVM = struct {
...
+        op1: ?*const relocatable.MaybeRelocatable,
+    ) !std.meta.Tuple(&[_]type{ ?relocatable.MaybeRelocatable, ?relocatable.MaybeRelocatable }) {
```
tcoratger, line 230: Here I think that `!std.meta.Tuple(&.{ ?MaybeRelocatable, ?MaybeRelocatable })` would work for something less verbose :)

src/vm/core.zig
```diff
@@ -213,6 +213,41 @@ pub const CairoVM = struct {
...
+        if (inst.opcode == .Call) {
```
tcoratger, line 231: Maybe here we can use a match as we are dealing with an enum (`Opcode`) in order not to forget any variant in the logic.

src/vm/core_test.zig
```diff
@@ -867,6 +870,200 @@ test "deduceOp1 when opcode == .AssertEq, res_logic == .Op1, no op0" {
     try expect(res.?.eq(relocatable.fromU64(7)));
 }
 
+test "deduceOp0 when opcode == .Call" {
+    // ************************************************************
+    // *                 SETUP TEST CONTEXT                       *
+    // ************************************************************
+    var vm = try CairoVM.init(gpa_allocator, .{});
```
tcoratger, line 877: In test I think this is better to use `std.testing.allocator` allocators which are designed for testing purposes no?

src/vm/core_test.zig
```diff
@@ -867,6 +870,200 @@ test "deduceOp1 when opcode == .AssertEq, res_logic == .Op1, no op0" {
...
+    const tuple = try vm.deduceOp0(&instr, null, null);
+    const op1 = tuple[0];
+    const res = tuple[1];
+
+    // ************************************************************
+    // *                      TEST CHECKS                         *
+    // ************************************************************
+    const expectedOp0: ?MaybeRelocatable = null; // temp var needed for type inference
+    const expectedRes: ?MaybeRelocatable = null;
+    try expectEqual(expectedOp0, op1);
```
tcoratger, line 947: Here if you want to be less verbose you can do:

```
try expectEqual(@as(?MaybeRelocatable, null), op1);
```

src/vm/core.zig
```diff
@@ -213,6 +213,41 @@ pub const CairoVM = struct {
...
+            var result: ?MaybeRelocatable = relocatable.newFromRelocatable(try self.run_context.pc.addUint(inst.size()));
+            return .{ result, null };
```
tcoratger, line 233: Here for something more compact, I think you can use:

```
return .{ relocatable.newFromRelocatable(try self.run_context.pc.addUint(inst.size())), null };
```

directly

src/vm/core_test.zig
abdelhamidbakhta: We should only use testing allocator yes. As it detects memory leaks and you can customize it’s behavior to reproduce some error cases

src/vm/core.zig
```diff
@@ -213,6 +213,41 @@ pub const CairoVM = struct {
...
+                return .{ try subOperands(dst.?.*, op1.?.*), dst.?.* };
+            } else if (dst != null and op1 != null and
+                dst.?.isFelt() and op1.?.isFelt() and
+                !op1.?.felt.isZero())
+            {
+                return .{
+                    relocatable.fromFelt(try dst.?.felt.div(op1.?.felt)),
+                    dst.?.*,
+                };
+            }
```
bingcicle, line 245: We're always going to check `dst` and `op1` in both these cases. We can extract some boolean comparisons here to make the if-else case easier to read:

```suggestion
            const dst_val = dst orelse return .{ null, null };
            const op1_val = op1 orelse return .{ null, null };
            if (inst.res_logic == .Add) {
                return .{
                    try subOperands(dst_val.*, op1_val.*),
                    dst_val.*,
                };
            } else if (dst_val.isFelt() and
                op1_val.isFelt() and
                !op1_val.felt.isZero())
            {
                return .{
                    relocatable.fromFelt(try dst_val.felt.div(op1_val.felt)),
                    dst_val.*,
                };
            }
```

Both are functionally equivalent, but I think less expressions in the conditionals make this easier to parse, and the conditions for early return (`dst == null` and `op1 == null`) are immediately obvious.

src/vm/core_test.zig
```diff
@@ -867,6 +870,200 @@ test "deduceOp1 when opcode == .AssertEq, res_logic == .Op1, no op0" {
...
+
+    const tuple = try vm.deduceOp0(&instr, null, null);
+    const op1 = tuple[0];
+    const res = tuple[1];
+
+    // ************************************************************
+    // *                      TEST CHECKS                         *
+    // ************************************************************
+    const expectedOp0: ?MaybeRelocatable = null; // temp var needed for type inference
+    const expectedRes: ?MaybeRelocatable = null;
```
bingcicle, line 946:
```suggestion
    const expected_op_0: ?MaybeRelocatable = null; // temp var needed for type inference
    const expected_res: ?MaybeRelocatable = null;
```

camelCase only for functions, snake_case for others.

src/vm/core.zig
lambda-0x: For most of styling related change i have followed #69, should i update both in this PR or open a new PR for deduce1 related things?

src/vm/core.zig
lambda-0x: with switch i would need to do something like:
```zig
        switch (inst.opcode) {
            .Call => {
                var result: ?MaybeRelocatable = relocatable.newFromRelocatable(try self.run_context.pc.addUint(inst.size()));
                return .{ result, null };
            },
            .AssertEq => {
                _ = dst orelse return .{ null, null };
                _ = op1 orelse return .{ null, null };
                if ((inst.res_logic == .Add)) {
                    return .{ try subOperands(dst.?.*, op1.?.*), dst.?.* };
                } else if (dst.?.isFelt() and op1.?.isFelt() and
                    !op1.?.felt.isZero())
                {
                    return .{
                        relocatable.fromFelt(try dst.?.felt.div(op1.?.felt)),
                        dst.?.*,
                    };
                }
            },
            .NOp => {
                return .{ null, null };
            },
            .Ret => {
                return .{ null, null };
            },
        }
        return .{ null, null };

```

Cannot use `_` (prong) for last 2 case because it can only be used on non exhaustive enums.

src/vm/core.zig
lambda-0x: agreed, would need to do something like this because methods are implemented on `T` not `*const T`
```zig
                _ = dst orelse return .{ null, null };
                _ = op1 orelse return .{ null, null };
                if ((inst.res_logic == .Add)) {
                    return .{ try subOperands(dst.?.*, op1.?.*), dst.?.* };
                } else if (dst.?.isFelt() and op1.?.isFelt() and
                    !op1.?.felt.isZero())
                {
                    return .{
                        relocatable.fromFelt(try dst.?.felt.div(op1.?.felt)),
                        dst.?.*,
                    };
                }

```

lambda-0x:
@tcoratger @bingcicle some of the suggestions you mentioned apply to deduceop1 as well so i will make those changes in this PR

lambda-0x:
some of suggestions mentioned by @tcoratger @bingcicle  apply to deduceop1 as well, should i make those changes in this PR or make a PR for it after this is merged? @lana-shanghai

src/vm/core_test.zig
lambda-0x: i think to be consistent with other tests making a variable makes more sense here

src/vm/core.zig
tcoratger: @lambda-0x A `else` would work no?

src/vm/core.zig
tcoratger: I think you can update everything in this PR because at the beginning of the file we declared `const MaybeRelocatable = relocatable.MaybeRelocatable;`. In my opinion, there is therefore no point in rewriting the entire `relocatable.MaybeRelocatable` each time. We can simply replace with `MaybeRelocatable` everywhere in this file if this is not done.

abdelhamidbakhta:
> some of suggestions mentioned by @tcoratger @bingcicle apply to deduceop1 as well, should i make those changes in this PR or make a PR for it after this is merged? @lana-shanghai

since those area of code are relatively close, you can proceed in the same PR @lambda-0x

src/vm/core.zig
bingcicle: Hmm, could you help me understand what you mean?

The suggestion has to do with the optional `?T` type, and not whether it's a const pointer or not. 

If this helps, think of `dst.?` as `dst.unwrap()` in Rust! Then, the suggested `orelse` line:

```zig
const dst_val = dst orelse return .{ null, null };
```
 is equivalent to something like:

```rust
let dst_val = maybe_dst.unwrap_or_else(|_| return (None, None))
```

src/vm/core.zig
lambda-0x: oh right, rust habits of using `_` in match statements :)

src/vm/core.zig
lambda-0x: got it

AbdelStark (approved):
lgtm

Issue #74, opened 2023-10-27
Add `memory` function as a test util function


tcoratger:
To improve and facilitate test coverage (e.g. for the following case which is currently under development: https://github.com/keep-starknet-strange/cairo-zig/issues/8), we need to start developing some useful tests for the core of the VM, the `memory` logic is one of them:
- Details of similar implementation in Rust can be found here: https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/utils.rs#L119-L128
- Unit tests to validate this implementation can be found here: https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/utils.rs#L628-L646
- This can be done after https://github.com/keep-starknet-strange/cairo-zig/issues/73 is resolved.

PR #89, opened 2023-10-31
feat: implement computeOp1Deductions

### LINKED ISSUE #68: feat: implement computeOp1Deductions
### Description
Once deducing op1 [issue](https://github.com/keep-starknet-strange/cairo-zig/issues/46) is complete, and `deduceMemoryCell` is [also](https://github.com/keep-starknet-strange/cairo-zig/issues/8) complete, we need to implement `computeOp1Deductions` which are called in `computeOperands`. It tries to deduce op_1 memory cell from builtins, if it fails (right now it returns `null` by default), it tries to deduce the operand from optional `dst` and `op_0`.

See Rust implementation [here](https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/vm/vm_core.rs#L524)

nils-mathieu:
Fixes #68 .

This PR intruduces the `conputeOp1Deductions` function.

# Blocked

This PR is blocked by the following issues: #74 #73 and #72 .

### DIFF
--- a/src/vm/core.zig
+++ b/src/vm/core.zig
@@ -206,7 +206,7 @@ pub const CairoVM = struct {
     /// - `dst`: The destination.
     /// - `op1`: The op1.
     pub fn computeOp0Deductions(
-        self: *Self,
+        self: *const Self,
         op_0_addr: MaybeRelocatable,
         instruction: *const instructions.Instruction,
         dst: ?MaybeRelocatable,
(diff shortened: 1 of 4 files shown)


src/vm/core.zig
```diff
@@ -219,6 +219,27 @@ pub const CairoVM = struct {
...
+    ) CairoVMError!MaybeRelocatable {
+        if (try self.deduceMemoryCell(op1_addr)) |op1| {
+            return op1;
+        } else {
+            const op1_deductions = deduceOp1(instruction, dst_op, op0);
+            if (op1_deductions[1]) |deduced_res| {
+                res.* = deduced_res;
+            }
+            const deducted_op1 = op1_deductions[0] orelse return CairoVMError.FailedToComputeOp1;
+            return deducted_op1;
```
tcoratger, line 239: I think that we can return directly without a previous variable declaration no?

src/vm/core.zig
```diff
@@ -219,6 +219,27 @@ pub const CairoVM = struct {
         _ = op_o;
     }
 
+    /// Runs deductions for op1.
```
abdelhamidbakhta, line 222: please add doc for Arguments and Returns

tcoratger:
For me you can add the following unit tests to validate the implementation:
- Fail to compute operands error if `op1_op` is `null`.
- Should return the result of `deduce_memory_cell` in normal situation where the result of `deduce_memory_cell` is not `null`. 
- If `deduce_memory_cell` operation is `null`, should return `op1`, the result of the `deduce_op1` operation and should also update `res` (if `null`) with the resulting `deduced_res`.
- Should return a virtual machine error if `deduce_op1` fails.

For me it's a good start to test the different uses of this function.

src/vm/core.zig
nils-mathieu: Yes but I think it's not as clear if `deducted_op1` never appears

PR #100, opened 2023-11-04
Range Check builtin

### LINKED ISSUE #5: feat: implement range check builtin
Copying a value to a range-checked cell allows to see if is inside a range `[0; 2**128)`.

See the [Go](https://github.com/NethermindEth/cairo-vm-go/blob/main/pkg/vm/builtins/range_check.go) and [Rust](https://github.com/lambdaclass/cairo-vm/blob/main/vm/src/vm/runners/builtin_runner/range_check.rs) implementations.

oxlime:
Fixes [#5 ](https://github.com/keep-starknet-strange/cairo-zig/issues/5)

need to add tests
formatting

### DIFF
--- a/src/vm/builtins/builtin_runner/range_check.zig
+++ b/src/vm/builtins/builtin_runner/range_check.zig
@@ -3,14 +3,15 @@ const Allocator = std.mem.Allocator;
 const ArrayList = std.ArrayList;
 
 const Felt252 = @import("../../../math/fields/starknet.zig").Felt252;
-const CELLS_PER_RANGE_CHECK = @import("../../types/range_check_instance_def.zig").CELLS_PER_RANGE_CHECK;
 const MemorySegmentManager = @import("../../memory/segments.zig").MemorySegmentManager;
 const relocatable = @import("../../memory/relocatable.zig");
 const Error = @import("../../error.zig");
 const validation_rule = @import("../../memory/memory.zig").validation_rule;
 const Memory = @import("../../memory/memory.zig").Memory;
 const Field = @import("../../../math/fields/starknet.zig").Field;
+const range_check_instance_def = @import("../../types/range_check_instance_def.zig");
 
+const CELLS_PER_RANGE_CHECK = range_check_instance_def.CELLS_PER_RANGE_CHECK;
 const Relocatable = relocatable.Relocatable;
 const MaybeRelocatable = relocatable.MaybeRelocatable;
 const MemoryError = Error.MemoryError;
@@ -107,7 +108,7 @@ pub const RangeCheckBuiltinRunner = struct {
     /// # Modifies
     /// - `self`: Updates the `base` value to the new segment's index.
     pub fn initializeSegments(self: *Self, segments: *MemorySegmentManager) void {
-        self.base = @as(usize, segments.addSegment().segment_index);
+        self.base = @intCast(segments.addSegment().segment_index);
     }
 
     /// Initializes and returns an `ArrayList` of `MaybeRelocatable` values.
--- a/src/vm/builtins/builtin_runner/range_check.zig
+++ b/src/vm/builtins/builtin_runner/range_check.zig
@@ -43,14 +43,16 @@ pub const RangeCheckBuiltinRunner = struct {
         n_parts: u32,
         included: bool,
     ) Self {
+        var bound = Felt252.one().shl(16 * n_parts);
+        var _bound: ?Felt252 = if (n_parts != 0 and bound.zero()) null else Felt252.new(bound);
+
         return .{
             .ratio = ratio,
             .base = 0,
             .stop_ptr = null,
             .cell_per_instance = range_check_instance_def.CELLS_PER_RANGE_CHECK,
             .n_input_cells = range_check_instance_def.CELLS_PER_RANGE_CHECK,
-            // TODO: implement shl logic: https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/vm/runners/builtin_runner/range_check.rs#L48-L53
-            ._bound = null,
+            ._bound = _bound,
             .included = included,
             .n_parts = n_parts,
             .instances_per_component = 1,
--- a/src/vm/builtins/builtin_runner/range_check.zig
+++ b/src/vm/builtins/builtin_runner/range_check.zig
@@ -1,5 +1,17 @@
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+const ArrayList = std.ArrayList;
+
 const Felt252 = @import("../../../math/fields/starknet.zig").Felt252;
 const range_check_instance_def = @import("../../types/range_check_instance_def.zig");
+const MemorySegmentManager = @import("../../memory/segments.zig").MemorySegmentManager;
+const relocatable = @import("../../memory/relocatable.zig");
+const Error = @import("../../error.zig");
+
+const Relocatable = relocatable.Relocatable;
+const MaybeRelocatable = relocatable.MaybeRelocatable;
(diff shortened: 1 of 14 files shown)


src/vm/builtins/builtin_runner/range_check.zig
```diff
@@ -43,26 +61,188 @@ pub const RangeCheckBuiltinRunner = struct {
         n_parts: u32,
         included: bool,
     ) Self {
+        var bound = Felt252.one().overflowing_shl(16 * n_parts);
+        var _bound: ?Felt252 = if (n_parts != 0 and bound.zero()) null else Felt252.new(bound);
```
tcoratger, line 65: `bound` and `_bound` could be declared as `const` instead of `var` because they are not modified afterwards. Maybe if this still is readable, for a more compact format, you can directly integrate those lines into the return statement without any variable declaration for a more compact format.

src/vm/builtins/builtin_runner/range_check.zig
```diff
@@ -43,26 +61,188 @@ pub const RangeCheckBuiltinRunner = struct {
...
+    /// This function adds a memory segment using the provided `segments` manager and
+    /// sets the `base` value to the index of the new segment.
+    ///
+    /// # Parameters
+    /// - `segments`: A pointer to the `MemorySegmentManager` for segment management.
+    ///
+    /// # Modifies
+    /// - `self`: Updates the `base` value to the new segment's index.
+    pub fn initialize_segments(self: *Self, segments: *MemorySegmentManager) void {
+        self.base = @as(usize, @intCast(segments.add().segment_index));
```
tcoratger, line 109: Here is `self.base = @intCast(segments.add().segment_index)` not sufficient as the type of `self.base` is already known in advance (type coercion)?

src/vm/builtins/builtin_runner/range_check.zig
oxlime: Will switch to const that makes sense.

As for the implementation the rust version is nearly identical.
 https://github.com/lambdaclass/cairo-vm/blob/0d41042b906258c28531797045d012a30b036ab1/vm/src/vm/runners/builtin_runner/range_check.rs#L48
I'm not sure that it would be readable if declared in the return as bound would be called twice in the assignment of _bound. It would be very long and personally think less readable.

tcoratger: Agreed

src/vm/builtins/builtin_runner/range_check.zig
```diff
@@ -43,26 +61,188 @@ pub const RangeCheckBuiltinRunner = struct {
...
     pub fn get_base(self: *const Self) usize {
         return self.base;
     }
+
+    /// Get the ratio value of this Range Check runner.
+    ///
+    /// # Returns
+    ///
+    /// The ratio value as an `u32`.
+    pub fn get_ratio(self: *const Self) ?u32 {
```
bingcicle, line 94:
```suggestion
    pub fn getRatio(self: *const Self) ?u32 {
```

> Unlike variables, which are snake_case, functions are camelCase.

source: https://ziglearn.org/chapter-1/#functions

Until the creator accepts the proposal to use snake_case for fns lets stick with camelCase for this (and the rest of your functions too).

oxlime: all of the builtin_runner structs are made with snake case get_base() so was following that formatting. 
https://github.com/keep-starknet-strange/cairo-zig/blob/6c0167b3c8711a5c79fae49f74c161a288531885/src/vm/builtins/builtin_runner/poseidon.zig#L68
Going to assume those are  a mistake. Will reformat.

tcoratger: @oxlime Yes, it was probably me who did that, it was indeed a mistake, I had forgotten this subtle change with Rust.

src/vm/builtins/builtin_runner/range_check.zig
```diff
@@ -43,26 +63,237 @@ pub const RangeCheckBuiltinRunner = struct {
...
+    /// An `ArrayList(Relocatable)` containing the rules address
+    /// verification fails.
+    pub fn rangeCheckValidationRule(memory: *Memory, address: Relocatable) std.ArrayList(!Relocatable) {
+        const addr = memory.get(address);
+        if (addr.bits <= N_PARTS * INNER_RC_BOUND_SHIFT) {
+            return std.ArrayList.append(address);
+        } else {
+            return std.ArrayList.append(Error.MemoryOutOfBounds);
+        }
+    }
```
tcoratger, line 256: I'm not sure this will work as expect because your `ArrayList` is not initialized before `append` method is used. See for example: https://github.com/ziglang/zig/blob/e74ced21b70f229167765d863613d3d5033b93fe/lib/std/array_list.zig#L1188-L1191

src/vm/builtins/builtin_runner/range_check.zig
```diff
@@ -43,26 +63,237 @@ pub const RangeCheckBuiltinRunner = struct {
...
+    /// # Parameters
+    ///
+    /// - `memory`: A `Memory` pointer of validation rules segment index.
+    /// - `address`: A `Relocatable` pointer to the validation rule.
+    ///
+    /// # Returns
+    ///
+    /// An `ArrayList(Relocatable)` containing the rules address
+    /// verification fails.
+    pub fn rangeCheckValidationRule(memory: *Memory, address: Relocatable) std.ArrayList(!Relocatable) {
```
tcoratger, line 249: Would say that the output should be `!std.ArrayList(Relocatable)` instead.

src/vm/builtins/builtin_runner/range_check.zig
```diff
@@ -43,26 +63,237 @@ pub const RangeCheckBuiltinRunner = struct {
...
+        const addr = memory.get(address);
+        if (addr.bits <= N_PARTS * INNER_RC_BOUND_SHIFT) {
+            return std.ArrayList.append(address);
```
tcoratger, line 252: We need a `try` to use the `append` method because this can return an error, see [here](https://github.com/ziglang/zig/blob/e74ced21b70f229167765d863613d3d5033b93fe/lib/std/array_list.zig#L266-L269) for a confirmation.

src/vm/types/range_check_instance_def.zig
```diff
@@ -3,10 +3,29 @@ pub const CELLS_PER_RANGE_CHECK: u32 = 1;
...
+
     /// Ratio
     ratio: ?u32,
     /// Number of 16-bit range checks that will be used for each instance of the builtin.
     ///
     /// For example, n_parts=8 defines the range [0, 2^128).
     n_parts: u32,
+
+    /// Number of cells per built in
+    pub fn cellsPerBuiltin() u32 {
```
tcoratger, line 16: Why do we need a function here as the `CELLS_PER_RANGE_CHECK` constant is declared as public as so already available? If it is still necessary to have this function, shouldn't we declare a constant in the structure which directly consumes this function at compile time to speed up the calculations? 

You have examples of that [here](https://kristoff.it/blog/what-is-zig-comptime/)

image

src/vm/types/range_check_instance_def.zig
```diff
@@ -3,10 +3,29 @@ pub const CELLS_PER_RANGE_CHECK: u32 = 1;
...
+        return CELLS_PER_RANGE_CHECK;
+    }
+
+    /// Number of units per builtin
+    pub fn rangeCheckUnitsPerBuiltin(self: *Self) u32 {
+        return self.n_parts;
+    }
+
+    pub fn default() !Self {
```
tcoratger, line 25: The return type can be `Self` instead of `!Self` because there is no potential error here no?

lana-shanghai (approved):
LGTM

PR #101, opened 2023-11-04
feat(vm): implement `updateRegisters` method and call it in inside `runInstruction` 

### LINKED ISSUE #10: feat: implement a function updateRegisters and call it in inside runInstruction
Right now there are functions to [update each register](https://github.com/keep-starknet-strange/cairo-zig/blob/main/src/vm/core.zig#L175). Create a function that updates all three registers based on an instruction, and make it callable inside `runInstruction` [function](https://github.com/keep-starknet-strange/cairo-zig/blob/main/src/vm/core.zig#L104).

tcoratger:
Should close #10.

### DIFF
--- a/src/vm/core.zig
+++ b/src/vm/core.zig
@@ -151,7 +151,11 @@ pub const CairoVM = struct {
         }
 
         const operands_result = try self.computeOperands(instruction);
-        _ = operands_result;
+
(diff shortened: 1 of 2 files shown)


lana-shanghai (approved):
LGTM

PR #112, opened 2023-11-07
feat: integrate all builtin runners in `deduceMemoryCell` VM function

### LINKED ISSUE #106: cleanup: return null if the deduce function is not implemented for a builtin
### Description
For example, the range check builtin does not have the `deduce` function, as we can see in the [python](https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/cairo/lang/builtins/bitwise/bitwise_builtin_runner.py) implementation, returns `None` in the [Rust](https://github.com/lambdaclass/cairo-vm/blob/main/vm/src/vm/runners/builtin_runner/range_check.rs#L107) implementation. But the builtin runner iterates over each builtin and [calls](https://github.com/keep-starknet-strange/cairo-zig/blob/b119ba4a2056441f76ed72bb2b6e3547cdc9b0e1/src/vm/core.zig#L437) the `deduce` function. Implement a function that returns `null` or treat the unimplemented path correctly. Make sure the runner iterates correctly over the builtins.

tcoratger:
This PR integrates all builtin runners within the `deduceMemoryCell` function of the `CairoVM` structure.
- A `deduceMemoryCell` function is integrated into the `BuiltinRunner` enum union which gives a switch between the different builtins.
- For now, as no builtin is implemented, all `deduceMemoryCell` functions return a `null`.
- The only existing implementation is the bitwise builtin but the implementation is currently in a separate file `/builtins/bitwise/bitwise.zig` awaiting migration to the `BitwiseBuiltinRunner` structure of the `builtins/builtin_runner.zig` file. So, at the moment a `BitwiseBuiltinRunnerTmp` type is imported and will have to be replaced by `BitwiseBuiltinRunner` once the implementation is migrated.

Should close #106.

### DIFF
--- a/src/vm/builtins/builtin_runner/bitwise.zig
+++ b/src/vm/builtins/builtin_runner/bitwise.zig
@@ -1,4 +1,7 @@
 const bitwise_instance_def = @import("../../types/bitwise_instance_def.zig");
+const Relocatable = @import("../../memory/relocatable.zig").Relocatable;
+const MaybeRelocatable = @import("../../memory/relocatable.zig").MaybeRelocatable;
+const Memory = @import("../../memory/memory.zig").Memory;
 
 /// Bitwise built-in runner
 pub const BitwiseBuiltinRunner = struct {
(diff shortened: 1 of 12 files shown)


lana-shanghai (approved):
LGTM

Issue #109, opened 2023-11-07
feat: add `MemoryCell` structure


tcoratger:
At the moment, in the `Memory` structure, the `data` field is a vector composed of elements of type `MaybeRelocatable`. 

https://github.com/keep-starknet-strange/cairo-zig/blob/747dc4a492d0f35e4894edb026f6273c7805d0a7/src/vm/memory/memory.zig#L31-L36

In the future, it may be desirable to add other information in addition to only a `MaybeRelocatable` type. In particular, I find the boolean flag `is_accessed` introduced by the Rust implementation quite useful in this context. I therefore propose the implementation of a `MemoryCell` structure which will initially contain the fields:
- `maybeRelocatable` of type `MaybeRelocatable`
- `isAccessed` of type boolean

This structure can then be extended and improved. In this same issue, it will also be necessary to replace the `MaybeRelocatable` type in the `data` vector of the `Memory` structure by the `MemoryCell` structure thus created.

Here are the Rust implementation details [here](https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/vm/vm_memory/memory.rs#L19-L42).

Deecellar:
Oh, seems interesting, I will take this feat

lana-shanghai:
> Oh, seems interesting, I will take this feat

Assigned!

lana-shanghai:
Hey @Deecellar could you give @oxlime a timeline when do you think this would be completed? His `runInstruction` PR is blocked by it

oxlime:
@lana-shanghai I can take this if @Deecellar isn't working on it anymore. Should be lots of new issues want to get pr through for fib program deadline.

oxlime:
have this almost finished I'm getting some errors in my Memory.markAsAccessed function that I think could be because of passing my value and not reference? Should i open a pr or can you check my local branch.  https://github.com/oxlime/cairo-zig/tree/feat/add_MemoryCell

Changes to Memory requires a lot of testing changes so just want to make sure this is the right direction.

PR #114, opened 2023-11-08
Feat/run instruction

### LINKED ISSUE #102: feat(vm): complete the implementation of `runInstruction`
At the moment, the implementation of the `runInstruction` method in the `CairoVM` structure is incomplete and needs to be completed and tested to work correctly:

https://github.com/keep-starknet-strange/cairo-zig/blob/6f683b4c98c75b5cf4ce36292f12ac3457619968/src/vm/core.zig#L138-L155

- Rust [reference implementation](https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/vm/vm_core.rs#L391-L432)
- Python [reference implementation](https://github.com/starkware-libs/cairo-lang/blob/9e6e0e96d208608d98635ccfad5b26285c4936e1/src/starkware/cairo/lang/vm/vm_core.py#L410-L433)

oxlime:
Run Instruction for vm core, fixes #102. 

Added rc limits as an optional tuple and counter step as well to the vm. 

#109 will need to be completed, Memory.markAsAccessed function used before completion.

### DIFF
--- a/src/vm/core.zig
+++ b/src/vm/core.zig
@@ -151,7 +151,11 @@ pub const CairoVM = struct {
         }
 
         const operands_result = try self.computeOperands(instruction);
-        _ = operands_result;
+
+        try self.updateRegisters(
+            instruction,
+            operands_result,
+        );
     }
 
     /// Compute the operands for a given instruction.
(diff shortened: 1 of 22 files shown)


src/vm/core.zig
```diff
@@ -156,6 +162,23 @@ pub const CairoVM = struct {
...
+        const OFFSET_BITS: u32 = 16;
+        const off_0 = instruction.off_0 + (@as(i16, 1) << (OFFSET_BITS - 1));
+        const off_1 = instruction.off_1 + (@as(i16, 1) << (OFFSET_BITS - 1));
+        const off_2 = instruction.off_2 + (@as(i16, 1) << (OFFSET_BITS - 1));
+
+        if (self.rc_limits == null) {
+            self.rc_limits = .{ @min(off_0, off_1, off_2), @max(off_0, off_1, off_2) };
+        } else {
+            self.rc_limits = .{ @min(self.rc_limits.?[0], off_0, off_1, off_2), @max(self.rc_limits.?[1], off_0, off_1, off_2) };
+        }
```
bingcicle, line 175: You can make use of [payload captures](https://ziglearn.org/chapter-1/#payload-captures) here, or `orelse`. if `off_0` is guaranteed we can do a little hack here:

```suggestion
        const limits = self.rc_limit orelse .{ off_0, off_0 };
        self.rc_limits = .{ @min(limits[0], off_0, off_1, off_2), @max(limits[1], off_0, off_1, off_2) };
```

The assumption here is `off_0` will always be around for this

PR #120, opened 2023-11-08
feat(vm): implement `computeOp0Deductions` method

### LINKED ISSUE #67: feat: implement computeOp0Deductions
### Description
Once deducing op0 [issue](https://github.com/keep-starknet-strange/cairo-zig/issues/45) is complete, and `deduceMemoryCell` is [also](https://github.com/keep-starknet-strange/cairo-zig/issues/8) complete, we need to implement `computeOp0Deductions` which are called in `computeOperands`. It tries to deduce op_0 memory cell from builtins, if it fails (right now it returns `null` by default), it tries to deduce the operand from optional `dst` and `op_1`.

See Rust implementation [here](https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/vm/vm_core.rs#L502)

tcoratger:
Should close #67 plus additional small refactoring.


lana-shanghai (approved):
lgtm

PR #127, opened 2023-11-10
fix: changed type defs in relocatable and added test

### LINKED ISSUE #115: feat: add `temp_data` to `Memory` structure
## Description:
To enable the addition of new temporary memory segments to the Memory Segment Manager, we need to introduce a `temp_data` field within the `Memory` structure. You can find an example of this implementation in the Rust code [here](https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/vm/vm_memory/memory.rs#L96). In order to resolve this issue effectively, the following additional implementations are required:

- **Add a Method for Creating Temporary Segments:** 
We need to introduce a method in the `MemorySegmentManager` structure that allows the creation of temporary segments. You can refer to the code example [here](https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/vm/vm_memory/memory_segments.rs#L43-L50).

- **Add a Method for Retrieving the Number of Temporary Segments:**
Another crucial addition is a method within the `MemorySegmentManager` structure that retrieves the count of segments in the temporary memory. You can find an example implementation for this task [here](https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/vm/vm_memory/memory_segments.rs#L29-L31).

This enhancement will improve the functionality of the Memory Segment Manager, enabling the management of temporary memory segments efficiently.

### LINKED ISSUE #117: feat: Modify `segment_index` type in `Relocatable` to `i64`
At the moment, `segment_index` type in `Relocatable` is `u64`. To add some temporary segments, we should be able to handle negative segment indexes:
- [Rust implementation](https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/vm/vm_memory/memory_segments.rs#L47)
- [Python implementation](https://github.com/starkware-libs/cairo-lang/blob/10d257de5f49543757599cad9afaa5b4dd6d752b/src/starkware/cairo/lang/vm/memory_segments.py#L54C7-L54C7)

As a consequence, the type of `segment_index` should be modified to `i64` in order to handle this use case.

Godspower-Eze:
Closes #117 and closes #115

### DIFF
--- a/src/vm/memory/memory.zig
+++ b/src/vm/memory/memory.zig
@@ -34,6 +34,13 @@ pub const Memory = struct {
         std.array_hash_map.AutoContext(Relocatable),
         true,
     ),
+    // The temporary data in the memory.
+    temp_data: std.ArrayHashMap(
+        Relocatable,
+        MaybeRelocatable,
+        std.array_hash_map.AutoContext(Relocatable),
+        true,
+    ),
     // The number of segments in the memory.
     num_segments: u32,
     // Validated addresses are addresses that have been validated.
@@ -69,6 +76,7 @@ pub const Memory = struct {
                 Relocatable,
                 MaybeRelocatable,
             ).init(allocator),
+            .temp_data = std.AutoArrayHashMap(Relocatable, MaybeRelocatable).init(allocator),
             .num_segments = 0,
             .validated_addresses = std.AutoHashMap(
                 Relocatable,
@@ -86,6 +94,7 @@ pub const Memory = struct {
     pub fn deinit(self: *Self) void {
         // Clear the hash maps
         self.data.deinit();
+        self.temp_data.deinit();
         self.validated_addresses.deinit();
         // Deallocate self.
         self.allocator.destroy(self);
@@ -108,19 +117,19 @@ pub const Memory = struct {
         MemoryOutOfBounds,
     }!void {
 
-        // Check if the address is valid.
-        if (address.segment_index < 0) {
-            return CairoVMError.InvalidMemoryAddress;
-        }
-
         // Insert the value into the memory.
-        self.data.put(
-            address,
-            value,
-        ) catch {
-            return CairoVMError.MemoryOutOfBounds;
-        };
-
+        if (address.segment_index < 0) {
+            self.temp_data.put(address, value) catch {
+                return CairoVMError.MemoryOutOfBounds;
+            };
+        } else {
+            self.data.put(
+                address,
+                value,
+            ) catch {
+                return CairoVMError.MemoryOutOfBounds;
+            };
+        }
         // TODO: Add all relevant checks.
     }
 
@@ -133,6 +142,9 @@ pub const Memory = struct {
         self: *Self,
         address: Relocatable,
     ) error{MemoryOutOfBounds}!MaybeRelocatable {
+        if (address.segment_index < 0) {
+            return self.temp_data.get(address) orelse CairoVMError.MemoryOutOfBounds;
+        }
         return self.data.get(address) orelse CairoVMError.MemoryOutOfBounds;
     }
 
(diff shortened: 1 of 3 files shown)


src/vm/memory/segments.zig
```diff
@@ -135,6 +149,15 @@ pub const MemorySegmentManager = struct {
         return self.memory.data.count();
     }
 
+    /// Retrieves the number of temporary memory segments.
+    ///
+    /// # Returns
+    ///
+    /// The number of temporary memory segments as a `usize`.
+    pub fn numTempSegments(self: *Self) usize {
```
tcoratger, line 157: Can we add a unit test for this function in order to be complete?

src/vm/memory/segments.zig
```diff
@@ -115,6 +115,20 @@ pub const MemorySegmentManager = struct {
...
     }
 
+    // Adds a temporary memory segment and returns the first address of the new segment.
+    pub fn addTempSegment(self: *Self) Relocatable {
+        // Increment the number of temporary segments.
+        self.memory.num_temp_segments += 1;
+
+        // Create the relocatable address for the new segment.
+        const relocatable_address = Relocatable{
+            .segment_index = -@as(i64, @intCast(self.memory.num_temp_segments)),
```
tcoratger, line 125: Here I think that as the type of `segment_index` in `Relocatable` is known by advance, `.segment_index = -@intCast(self.memory.num_temp_segments)` should be enough no?

Godspower-Eze: The `num_temp_segments` is a `u32` by default so this `-@intCast(self.memory.num_temp_segments)` doesn't work in this case.

tcoratger: @Godspower-Eze Ah yes true :)

tcoratger (approved):

src/vm/memory/segments.zig
Godspower-Eze: Done!

it's not a good pattern to add to the hash map directly using `put` because nothing stops me from adding a segment with a negative index to the `data` and not `temp_data`.

tcoratger:
lgtm

src/vm/memory/memory.zig
```diff
@@ -108,19 +120,19 @@ pub const Memory = struct {
         MemoryOutOfBounds,
     }!void {
 
-        // Check if the address is valid.
+        // Insert the value into the memory.
         if (address.segment_index < 0) {
-            return CairoVMError.InvalidMemoryAddress;
+            self.temp_data.put(address, value) catch {
+                return CairoVMError.MemoryOutOfBounds;
```
lana-shanghai, line 126: Do we need the `InvalidMemoryAddress` error or we put these under the same umbrella of `MemoryOutOfBounds`?

lana-shanghai: If the error isn't used anymore remove it from `vm/error.zig`

tcoratger: I saw this earlier but I said to myself that we would put it in https://github.com/keep-starknet-strange/cairo-zig/issues/124 because the `set` function anyway here is not complete and therefore cannot be tested properly.

Indeed, here we always catch the put error in a `MemoryOutOfBounds` but I think that we must first add all the tests and all the verifications properly as in the Python or Rust implementation before being able to rule on this problem which is not as simple as that in my opinion because before `put` we must first do a certain number of checks with the potential return of errors.

lana-shanghai: OK fair enough, I'll approve then, let's review the used and unused errors at the end of first milestone

PR #129, opened 2023-11-10
feat: implement `computeOp1Deductions` method in `CairoVM` structure

### LINKED ISSUE #68: feat: implement computeOp1Deductions

tcoratger:
Should close #68.

### DIFF
--- a/src/vm/core.zig
+++ b/src/vm/core.zig
@@ -344,6 +344,39 @@ pub const CairoVM = struct {
         return op0_op orelse CairoVMError.FailedToComputeOperands;
     }
 
+    /// Compute Op1 deductions based on the provided instruction, destination, and Op0.
+    ///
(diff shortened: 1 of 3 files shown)


lana-shanghai (approved):
lgtm

PR #91, opened 2023-10-31
add `cmp` utils for `Felt252`

tcoratger:
This PRs adds some util primitives for field element comparisons with unit tests.

PR #132, opened 2023-11-11
feat: fix comparison for `Felt252` and unit tests

tcoratger:
This PR adds a fix related to #91. 

In the `Felt252` structure, the field elements are represented in Montgomery form to facilitate multiplications. So, when comparing two `Felt252`, we must first convert to non-Montgomery format before comparing the limbs. This is what this PR achieves with the appropriate unit tests which allow this correction to be validated.


lana-shanghai (approved):
LGTM

PR #133, opened 2023-11-11
Add/memory inner

### LINKED ISSUE #72: Add `memory_inner` function as a test util function
To improve and facilitate test coverage (e.g. for the following case which is currently under development: https://github.com/keep-starknet-strange/cairo-zig/issues/8), we need to start developing some useful tests for the core of the VM, the `memory_inner` logic is one of them:
- Details of similar implementation in Rust can be found here: https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/utils.rs#L141-L173

oxlime:
Closes #72

### DIFF
--- a/src/vm/core.zig
+++ b/src/vm/core.zig
@@ -151,7 +151,11 @@ pub const CairoVM = struct {
         }
 
         const operands_result = try self.computeOperands(instruction);
-        _ = operands_result;
+
+        try self.updateRegisters(
+            instruction,
+            operands_result,
+        );
     }
 
     /// Compute the operands for a given instruction.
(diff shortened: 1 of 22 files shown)


src/vm/core.zig
```diff
@@ -40,6 +40,10 @@ pub const CairoVM = struct {
     is_run_finished: bool,
     /// VM trace
     trace_context: TraceContext,
+    /// Current Step
```
tcoratger, line 43: Are these changes in the core file normal? Maybe the branch is not up to date with main?

src/vm/memory/memory.zig
```diff
@@ -207,6 +207,69 @@ pub const Memory = struct {
...
 
+// Utility function to help set up memory for tests
+//
+// # Arguments
+// - `memory` - memory to be set
+// - `vals` - complile time structure with heterogenous types
+fn memoryInner(memory: *Memory, comptime vals: anytype) !void {
+    inline for (vals) |row| {
+        const firstCol = row[0];
+        const relo = Relocatable.new(firstCol[0], firstCol[1]);
```
tcoratger, line 218: Small detail but the proper name for this is `const address = Relocatable.new(firstCol[0], firstCol[1]);` because this corresponds to the address in memory no?

src/vm/memory/memory.zig
```diff
@@ -207,6 +207,69 @@ pub const Memory = struct {
...
+        // Check number of inputs in row
+        if (row[1].len == 1) {
+            const nextCol = row[1];
+            const maybeFromFelt = MaybeRelocatable{ .felt = Felt252.fromInteger(nextCol[0]) };
+            try memory.set(relo, maybeFromFelt);
+        } else {
+            const val = row[1];
```
tcoratger, line 225: As this approach is used both here (`const val = row[1];`) and also in the main if (`const nextCol = row[1];`) perhaps we could declare the constant in a general outside of the if, before no?

src/vm/memory/memory.zig
```diff
@@ -207,6 +207,69 @@ pub const Memory = struct {
...
+// - `vals` - complile time structure with heterogenous types
+fn memoryInner(memory: *Memory, comptime vals: anytype) !void {
+    inline for (vals) |row| {
+        const firstCol = row[0];
+        const relo = Relocatable.new(firstCol[0], firstCol[1]);
+        // Check number of inputs in row
+        if (row[1].len == 1) {
+            const nextCol = row[1];
+            const maybeFromFelt = MaybeRelocatable{ .felt = Felt252.fromInteger(nextCol[0]) };
+            try memory.set(relo, maybeFromFelt);
```
tcoratger, line 223: Here, it's a preference that I have to make the lines compact, it's whatever you prefer, I think that the three lines above can be reduced to:

```
try memory.set(relo, .{ .felt = Felt252.fromInteger(row[1][0]) });
```

src/vm/memory/memory.zig
```diff
@@ -207,6 +207,69 @@ pub const Memory = struct {
...
+                .Pointer => {
+                    const num = try std.fmt.parseUnsigned(i64, val[0], 10);
+                    const strRelo = Relocatable.new(num, val[1]);
+                    const maybeFromString = MaybeRelocatable{ .relocatable = strRelo };
+                    try memory.set(relo, maybeFromString);
+                },
+                else => {
+                    const second = Relocatable.new(val[0], val[1]);
+                    const maybeSec = MaybeRelocatable{ .relocatable = second };
+                    try memory.set(relo, maybeSec);
```
tcoratger, line 237: Again my preference for line compacity :), we could have:

```
try memory.set(relo, .{ .relocatable = Relocatable.new(val[0], val[1])});
```

src/vm/memory/memory.zig
```diff
@@ -207,6 +207,69 @@ pub const Memory = struct {
...
+            try memory.set(relo, maybeFromFelt);
+        } else {
+            const val = row[1];
+            const T = @TypeOf(val[0]);
+            switch (@typeInfo(T)) {
+                .Pointer => {
+                    const num = try std.fmt.parseUnsigned(i64, val[0], 10);
+                    const strRelo = Relocatable.new(num, val[1]);
+                    const maybeFromString = MaybeRelocatable{ .relocatable = strRelo };
+                    try memory.set(relo, maybeFromString);
```
tcoratger, line 232: Maybe this one is more readable with two lines but we could have:
```
const num = try std.fmt.parseUnsigned(i64, val[0], 10);
try memory.set(relo, .{ .relocatable = Relocatable.new(num, val[1]) });
```

src/vm/core.zig
oxlime: think i accidently mixed this pr with my other pending pr will fix

src/vm/memory/memory.zig
oxlime: Yes I like this is much better not sure I was able to access row[1][0] like that but since the change to a const row[1] outside of if statement its no longer an concern.

src/vm/memory/memory.zig
oxlime: agreed again

tcoratger (approved):

tcoratger:
lgtm

PR #134, opened 2023-11-11
feat: add comparison utils in `Relocatable` and `MaybeRelocatable`

### LINKED ISSUE #126: feat: add comparison utils in `Relocatable` and `MaybeRelocatable`
Currently we have some basic equality, addition and subtraction operations in the `Relocatable` structure and in the `MaybeRelocatable` union enum. However, for each of these two data structures we are missing the following comparison tools:
- Less than (between two `Relocatable` and then between two `MaybeRelocatable`)
- Less or equal than (between two `Relocatable` and then between two `MaybeRelocatable`)
- Greater than (between two `Relocatable` and then between two `MaybeRelocatable`)
- Greater or equal than (between two `Relocatable` and then between two `MaybeRelocatable`)

**Reference - Python implementation:** https://github.com/starkware-libs/cairo-lang/blob/10d257de5f49543757599cad9afaa5b4dd6d752b/src/starkware/cairo/lang/vm/relocatable.py#L46C5-L46C5

0xicosahedron:
Fix: #126

### DIFF
--- a/src/vm/memory/relocatable.zig
+++ b/src/vm/memory/relocatable.zig
@@ -51,7 +51,7 @@ pub const Relocatable = struct {
         self: Self,
         other: Self,
     ) bool {
-        return self.segment_index < other.segment_index and self.offset < other.offset;
+        return self.segment_index < other.segment_index or (self.segment_index == other.segment_index and self.offset < other.offset);
     }
 
     // Determines if this Relocatable is less than or equal to another.
@@ -63,7 +63,7 @@ pub const Relocatable = struct {
         self: Self,
         other: Self,
     ) bool {
-        return self.lt(other) or self.eq(other);
+        return self.segment_index < other.segment_index or (self.segment_index == other.segment_index and self.offset <= other.offset);
     }
 
     // Determines if this Relocatable is greater than another.
@@ -75,7 +75,7 @@ pub const Relocatable = struct {
         self: Self,
         other: Self,
     ) bool {
-        return !self.lt(other);
+        return self.segment_index > other.segment_index or (self.segment_index == other.segment_index and self.offset > other.offset);
     }
 
     // Determines if this Relocatable is greater than or equal to another.
@@ -87,7 +87,7 @@ pub const Relocatable = struct {
         self: Self,
         other: Self,
     ) bool {
-        return !self.le(other);
+        return self.segment_index > other.segment_index or (self.segment_index == other.segment_index and self.offset >= other.offset);
     }
 
     /// Attempts to subtract a `Relocatable` from another.
@@ -610,6 +610,42 @@ test "Relocatable: addMaybeRelocatableInplace should return an error if other is
     );
 }
 
+test "Relocatable: lt should return true if other relocatable is greater than or equal, false otherwise" {
+    try expect(Relocatable.new(-1, 2).lt(Relocatable.new(-1, 3)));
+    try expect(Relocatable.new(1, 5).lt(Relocatable.new(2, 4)));
+
+    try expect(!Relocatable.new(2, 4).lt(Relocatable.new(2, 4)));
+    try expect(!Relocatable.new(2, 5).lt(Relocatable.new(2, 4)));
+    try expect(!Relocatable.new(3, 3).lt(Relocatable.new(2, 4)));
+}
+
+test "Relocatable: le should return true if other relocatable is greater, false otherwise" {
+    try expect(Relocatable.new(-1, 2).le(Relocatable.new(-1, 3)));
+    try expect(Relocatable.new(1, 5).le(Relocatable.new(2, 4)));
+    try expect(Relocatable.new(2, 4).le(Relocatable.new(2, 4)));
+
+    try expect(!Relocatable.new(2, 5).le(Relocatable.new(2, 4)));
+    try expect(!Relocatable.new(3, 3).le(Relocatable.new(2, 4)));
+}
+
+test "Relocatable: gt should return true if other relocatable is less than or equal, false otherwise" {
+    try expect(!Relocatable.new(-1, 2).gt(Relocatable.new(-1, 3)));
+    try expect(!Relocatable.new(1, 5).gt(Relocatable.new(2, 4)));
+    try expect(!Relocatable.new(2, 4).gt(Relocatable.new(2, 4)));
+
+    try expect(Relocatable.new(2, 5).gt(Relocatable.new(2, 4)));
+    try expect(Relocatable.new(3, 3).gt(Relocatable.new(2, 4)));
+}
+
+test "Relocatable: ge should return true if other relocatable is less, false otherwise" {
+    try expect(!Relocatable.new(-1, 2).gt(Relocatable.new(-1, 3)));
+    try expect(!Relocatable.new(1, 5).gt(Relocatable.new(2, 4)));
+    try expect(!Relocatable.new(2, 4).gt(Relocatable.new(2, 4)));
+
+    try expect(Relocatable.new(2, 5).gt(Relocatable.new(2, 4)));
+    try expect(Relocatable.new(3, 3).gt(Relocatable.new(2, 4)));
+}
+
 test "MaybeRelocatable: eq should return true if two MaybeRelocatable are the same (Relocatable)" {
     var maybeRelocatable1 = MaybeRelocatable{ .relocatable = Relocatable.new(0, 10) };
     var maybeRelocatable2 = MaybeRelocatable{ .relocatable = Relocatable.new(0, 10) };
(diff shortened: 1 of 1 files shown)


src/vm/memory/relocatable.zig
```diff
@@ -42,6 +42,54 @@ pub const Relocatable = struct {
...
+    // Determines if this Relocatable is greater than another.
+    // # Arguments
+    // - other: The other Relocatable to compare to.
+    // # Returns
+    // `true` if self is greater than other, `false` otherwise.
+    pub fn gt(
+        self: Self,
+        other: Self,
+    ) bool {
+        return !self.lt(other);
```
tcoratger, line 78: Here I think it doesn't work because if `self` and `other` are equal, then `!self.lt(other)` is true and yet `self` is not greater than `other`. So I think you either have to do `!self.le(other)` or write the comparison in its entirety `self.segment_index > other.segment_index and self.offset > other.offset` which is more efficient from a performance point of view because the code does not need to run both `self.lt(other)` then `self.eq(other)` and then do the negation in the case of equality.

src/vm/memory/relocatable.zig
```diff
@@ -42,6 +42,54 @@ pub const Relocatable = struct {
...
+    // Determines if this Relocatable is greater than or equal to another.
+    // # Arguments
+    // - other: The other Relocatable to compare to.
+    // # Returns
+    // `true` if self is greater than or equal to other, `false` otherwise.
+    pub fn ge(
+        self: Self,
+        other: Self,
+    ) bool {
+        return !self.le(other);
```
tcoratger, line 90: In this situation it is the same as for the previous example. If `self` and `other` are equal the assertion `!self.le(other)` is not true and yet `self` is greater than or equal to `other` because the two instances are equal.

src/vm/memory/relocatable.zig
```diff
@@ -42,6 +42,54 @@ pub const Relocatable = struct {
         return self.segment_index == other.segment_index and self.offset == other.offset;
     }
 
+    // Determines if this Relocatable is less than another.
+    // # Arguments
+    // - other: The other Relocatable to compare to.
+    // # Returns
+    // `true` if self is less than other, `false` otherwise.
+    pub fn lt(
+        self: Self,
```
tcoratger, line 51: Finally after our discussion you didn't use pointers here?

tcoratger:
I've added a few small comments here. After unit tests this seems good to me.

src/vm/memory/relocatable.zig
0xicosahedron: yeah typo on my side, its same as you suggested in the python implementation.

src/vm/memory/relocatable.zig
0xicosahedron: there are other places (like `eql` in `time.zig`, `equal` in `field.zig`) where we take `Self` as well so for consistency i kept it the same. 

if taking a pointer instead of `Self` helps in performance then we would update it everywhere but i am not sure if thats the case.

tcoratger: Ok then let's keep it like this because it is the objective of the PR I think to change this formalism if it is used anywhere else in the codebase in this way. Let's possibly open another issue that would change the values for pointers in comparisons of this type everywhere in the code, what do you think @lana-shanghai?

0xicosahedron:
added tests, please rereview @tcoratger

src/vm/memory/relocatable.zig
```diff
@@ -438,6 +610,42 @@ test "Relocatable: addMaybeRelocatableInplace should return an error if other is
     );
 }
 
+test "Relocatable: lt should return true if other relocatable is greater than or equal, false otherwise" {
+    try expect(Relocatable.new(-1, 2).lt(Relocatable.new(-1, 3)));
+    try expect(Relocatable.new(1, 5).lt(Relocatable.new(2, 4)));
+
+    try expect(!Relocatable.new(2, 4).lt(Relocatable.new(2, 4)));
+    try expect(!Relocatable.new(2, 5).lt(Relocatable.new(2, 4)));
```
tcoratger, line 618: Can we add also a test for `Relocatable.new(2, 4).lt(Relocatable.new(2, 5));` I think it represents the second condition after the `or` inside your `lt` function?

src/vm/memory/relocatable.zig
```diff
@@ -438,6 +610,42 @@ test "Relocatable: addMaybeRelocatableInplace should return an error if other is
...
+    try expect(!Relocatable.new(3, 3).lt(Relocatable.new(2, 4)));
+}
+
+test "Relocatable: le should return true if other relocatable is greater, false otherwise" {
+    try expect(Relocatable.new(-1, 2).le(Relocatable.new(-1, 3)));
+    try expect(Relocatable.new(1, 5).le(Relocatable.new(2, 4)));
+    try expect(Relocatable.new(2, 4).le(Relocatable.new(2, 4)));
+
+    try expect(!Relocatable.new(2, 5).le(Relocatable.new(2, 4)));
```
tcoratger, line 627: Can we add also a test for `Relocatable.new(2, 4).le(Relocatable.new(2, 5));` ?

src/vm/memory/relocatable.zig
```diff
@@ -438,6 +610,42 @@ test "Relocatable: addMaybeRelocatableInplace should return an error if other is
...
+    try expect(!Relocatable.new(3, 3).le(Relocatable.new(2, 4)));
+}
+
+test "Relocatable: gt should return true if other relocatable is less than or equal, false otherwise" {
+    try expect(!Relocatable.new(-1, 2).gt(Relocatable.new(-1, 3)));
+    try expect(!Relocatable.new(1, 5).gt(Relocatable.new(2, 4)));
+    try expect(!Relocatable.new(2, 4).gt(Relocatable.new(2, 4)));
+
+    try expect(Relocatable.new(2, 5).gt(Relocatable.new(2, 4)));
```
tcoratger, line 636: Can we add also a test for `Relocatable.new(2, 4).gt(Relocatable.new(2, 5));` ?

src/vm/memory/relocatable.zig
```diff
@@ -438,6 +610,42 @@ test "Relocatable: addMaybeRelocatableInplace should return an error if other is
...
+    try expect(Relocatable.new(3, 3).gt(Relocatable.new(2, 4)));
+}
+
+test "Relocatable: ge should return true if other relocatable is less, false otherwise" {
+    try expect(!Relocatable.new(-1, 2).gt(Relocatable.new(-1, 3)));
+    try expect(!Relocatable.new(1, 5).gt(Relocatable.new(2, 4)));
+    try expect(!Relocatable.new(2, 4).gt(Relocatable.new(2, 4)));
+
+    try expect(Relocatable.new(2, 5).gt(Relocatable.new(2, 4)));
```
tcoratger, line 645: Can we add also a test for `Relocatable.new(2, 4).ge(Relocatable.new(2, 5));` ?

src/vm/memory/relocatable.zig
```diff
@@ -438,6 +610,42 @@ test "Relocatable: addMaybeRelocatableInplace should return an error if other is
...
+    try expect(Relocatable.new(3, 3).gt(Relocatable.new(2, 4)));
+}
+
+test "Relocatable: ge should return true if other relocatable is less, false otherwise" {
+    try expect(!Relocatable.new(-1, 2).gt(Relocatable.new(-1, 3)));
```
tcoratger, line 641: Here I think that you wanted to test `ge` instead of `gt` no?

src/vm/memory/relocatable.zig
```diff
@@ -537,6 +745,26 @@ test "MaybeRelocatable: tryIntoU64 should return an error if MaybeRelocatable Fe
...
+test "MaybeRelocatable: any comparision should return false if other MaybeRelocatable is of different variant 2" {
+    var maybeRelocatable1 = MaybeRelocatable{ .felt = Felt252.fromInteger(10) };
+    var maybeRelocatable2 = MaybeRelocatable{ .relocatable = Relocatable.new(0, 10) };
+
+    try expect(!maybeRelocatable1.lt(maybeRelocatable2));
+    try expect(!maybeRelocatable1.le(maybeRelocatable2));
+    try expect(!maybeRelocatable1.gt(maybeRelocatable2));
+    try expect(!maybeRelocatable1.ge(maybeRelocatable2));
+}
+
```
tcoratger, line 767: Can we also test one true assertion for each `MaybeRelocatable` comparison case? Like:
- One `lt` between two `Relocatable`
- One `le` between two `Relocatable`
- One `gt` between two `Relocatable`
- One `ge` between two `Relocatable`
- One `lt` between two `Felt252`
- One `le` between two `Felt252`
- One `gt` between two `Felt252`
- One `ge` between two `Felt252`

src/vm/memory/relocatable.zig
0xicosahedron: my bad, copy pasted the test and forgot to update the body xP

src/vm/memory/relocatable.zig
0xicosahedron: i thought about it but since when both variants are same, result is same as operation on that type directly so its already covered by other tests.

src/vm/memory/relocatable.zig
0xicosahedron: that is tested by 
>   Relocatable.new(-1, 2).lt(Relocatable.new(-1, 3))

src/vm/memory/relocatable.zig
tcoratger: Yes I know it's just to confirm that for example there is not a hidden typo in one of the variants which would call `ge` instead of `gt` or `lt`. But in absolute terms it's not a necessity. If it takes you a few minutes do it otherwise we'll do it later when we don't know what to do :)

src/vm/memory/relocatable.zig
tcoratger: Yes ok, sounds good then

src/vm/memory/relocatable.zig
0xicosahedron: > Yes I know it's just to confirm that for example there is not a hidden typo
right that make sense, i will add them

src/vm/memory/relocatable.zig
```diff
@@ -564,3 +900,11 @@ test "fromU64: should create a MaybeRelocatable from a u64" {
...
     );
 }
+
+fn maybeRelocatableRelocatable(segment_index: i64, offset: u64) MaybeRelocatable {
+    return MaybeRelocatable{ .relocatable = Relocatable.new(segment_index, offset) };
+}
+
+fn maybeRelocatableFelt(num: u256) MaybeRelocatable {
+    return MaybeRelocatable{ .felt = Felt252.fromInteger(num) };
+}
```
0xicosahedron, line 914: make this helper function because declaring variables was getting very hard, but this name it a bit confusing, suggestions welcome.

tcoratger: @0xicosahedron I think that:
- `maybeRelocatableFelt` already exists with 
https://github.com/keep-starknet-strange/cairo-zig/blob/35b888f6bd2496b0c80572f77bd3e2c3fdef9097/src/vm/memory/relocatable.zig#L273-L275
- For `maybeRelocatableRelocatable` you can simplify it by making use of:
https://github.com/keep-starknet-strange/cairo-zig/blob/35b888f6bd2496b0c80572f77bd3e2c3fdef9097/src/vm/memory/relocatable.zig#L255-L257

I mean something like:
```zig
fromSegment(segment_index: i64, offset: u64) MaybeRelocatable {
     return newFromRelocatable(Relocatable.new(segment_index, offset));
}
```

tcoratger: Pending this one otherwise looks good to me

0xicosahedron: got it, removed felt variant and hoisted relocatable variant to be near `fromU256` with documentation

tcoratger (approved):

PR #142, opened 2023-11-13
Li/feat/compute operands

lana-shanghai:


src/vm/core.zig
```diff
@@ -282,37 +282,56 @@ pub const CairoVM = struct {
...
-            op_0_op,
+            op_res.op_0,
         );
-        const op_1_op = try self.segments.memory.get(op_1_addr);
 
-        const res = try computeRes(instruction, op_0_op, op_1_op);
+        op_res.op_1 = try self.segments.memory.get(op_res.op_1_addr);
+
+        op_res.res = try computeRes(instruction, op_res.op_0, op_res.op_1);
+        const res_op: ?MaybeRelocatable = op_res.res;
```
tcoratger, line 301: I'm not sure but here you compute directly `computeRes` and you assign the result to the variable `res`. But for me what is done in Rust is a little different:
- The `res` variable is first initialized to `null`.
- `compute_op0_deductions` and `compute_op1_deductions` are calculated (if `op0_op` or `op1_op` are zero respectively, which is equivalent to an error for us), which can possibly modify the `res`.
- Then after these calculations, if `res` is still `null` then we calculate `compute_res` but not otherwise, which can lead to slightly different results I think.

src/vm/core.zig
```diff
@@ -282,37 +282,56 @@ pub const CairoVM = struct {
...
+
+        op_res.op_0 = self.segments.memory.get(op_res.op_0_addr) catch
+            try self.computeOp0Deductions(
+                op_res.op_0_addr,
+                instruction,
+                dst_ptr,
+                op_1_ptr,
+            );
+
+        op_res.op_1 = self.segments.memory.get(op_res.op_1_addr) catch
```
tcoratger, line 317: Same comment as for `op_0` here due to the lines:

```zig
op_res.op_1_addr = try self.run_context.computeOp1Addr(instruction, op_res.op_0);
op_res.op_1 = try self.segments.memory.get(op_res.op_1_addr);
```
that can give rise to an error if `self.segments.memory.get(op_res.op_1_addr)` gives an error no? So I would perhaps recommend putting `catch` patterns directly in the assignments of `op_res.op_0` and `op_res.op_1` above, no?

lana-shanghai: If you do a `catch` at `op_res.op_1 = try self.segments.memory.get(op_res.op_1_addr);` then you need to do `computeOp1Deductions` and this function needs `res` already computed. That's why I divided the first computation of `op_1` and the second one with the `catch`. 

Or you mean that in the first `op_1` we need to do 

```
op_res.op_1 = self.segments.memory.get(op_res.op_1_addr) catch null;
``` 
to assign `null` to `op_1` from the beginning?

tcoratger: Yes that would be correct I think to avoid being stop there due to an error but assigning a null instead

src/vm/core.zig
lana-shanghai: You are correct I jumbled up the moment of assigning `res`

tcoratger (approved):

PR #143, opened 2023-11-13
feat: add `relocation_rules` field in `Memory` structure

### LINKED ISSUE #140: feat: add `relocation_rules` field in `Memory` structure
We need to add a HashMap of relocation rules in order to be able to relocate memory according to the given relocation rules. So it should be good to add the `relocation_rules` HashMap field in the `Memory` structure as it is done in the Rust implementation:

https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/vm/vm_memory/memory.rs#L97-L99

and also to add a method to fill this HashMap as done here:

https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/vm/vm_memory/memory.rs#L245-L268

tcoratger:
Should close #140.

### DIFF
--- a/src/vm/error.zig
+++ b/src/vm/error.zig
@@ -17,6 +17,9 @@ pub const CairoVMError = error{
 
 pub const MemoryError = error{
     MissingSegmentUsedSizes,
+    AddressNotInTemporarySegment,
+    NonZeroOffset,
+    DuplicatedRelocation,
 };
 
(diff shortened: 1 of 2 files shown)


lana-shanghai (approved):
lgtm

PR #145, opened 2023-11-13
refactor: small refactor and more explicit Ap update

tcoratger:
This PR adds a bit of refactoring and a more explicit Ap update by clearly mentioning all the variants of `ApUpdate` enum instead of using `else` statement.


lana-shanghai (approved):
lgtm

PR #146, opened 2023-11-14
Add/util segments

### LINKED ISSUE #75: Add `segments` function as a test util function
To improve and facilitate test coverage (e.g. for the following case which is currently under development: https://github.com/keep-starknet-strange/cairo-zig/issues/8), we need to start developing some useful tests for the core of the VM, the `segments` logic is one of them:
- Details of similar implementation in Rust can be found here: https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/utils.rs#L102-L117
- This can be done after #74 is resolved.

oxlime:
Closes #75 uses same interface as #139.

### DIFF
--- a/src/vm/memory/segments.zig
+++ b/src/vm/memory/segments.zig
@@ -8,6 +8,7 @@ const expectError = std.testing.expectError;
 
 // Local imports.
 const Memory = @import("memory.zig").Memory;
+const MemoryCell = @import("memory.zig").MemoryCell;
 const relocatable = @import("relocatable.zig");
 const Relocatable = @import("relocatable.zig").Relocatable;
 const MaybeRelocatable = @import("relocatable.zig").MaybeRelocatable;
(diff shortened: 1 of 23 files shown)


src/vm/memory/segments.zig
```diff
@@ -219,6 +220,15 @@ pub const MemorySegmentManager = struct {
     }
 };
 
+// Utility function to help set up memory segments
+//
+// # Arguments
+// - `segment_manager` - MemorySegmentManger to be passed in
+// - `vals` - complile time structure with heterogenous types
+pub fn segmentsUtil(segment_manager: *MemorySegmentManager, comptime vals: anytype) !void {
+    try memoryFile.memoryInner(segment_manager.memory, vals);
```
lana-shanghai, line 229: there was a PR with renaming `memoryInner` to `setUpMemory` you will need to change this line

src/vm/memory/segments.zig
```diff
@@ -525,3 +535,21 @@ test "MemorySegmentManager: isValidMemoryValue should return true if valid segme
     var value: MaybeRelocatable = .{ .relocatable = Relocatable.new(0, 5) };
```
lana-shanghai, line 535: Could we use here also `fromSegment` like in the `core_test` file? Looks much neater

oxlime: Yes agree much cleaner

lana-shanghai (approved):
lgtm

PR #149, opened 2023-11-14
doc: documentation for Cairo Error Types

ndcroos:
See issue https://github.com/keep-starknet-strange/cairo-zig/issues/137

### DIFF
--- a/src/vm/error.zig
+++ b/src/vm/error.zig
@@ -18,7 +18,7 @@ pub const CairoVMError = error{
     TypeMismatchNotFelt,
     /// Error encountered with a built-in runner.
     RunnerError,
-    /// Occurs when the expected value is not a Relocatable (e.g. not a Felt),
+    /// Occurs when the expected value is not a Relocatable,
     /// or when subtracting two relocatables with different segment indices.
     TypeMismatchNotRelocatable,
     /// Value is too large to be coerced to a u64.
--- a/src/vm/error.zig
+++ b/src/vm/error.zig
@@ -12,7 +12,7 @@ pub const CairoVMError = error{
     InstructionFetchingFailed,
     /// Error in converting the encoded instruction to a u64.
     InstructionEncodingError,
-    /// TODO, Logic error during parsing of the resolution logic?
+    /// TODO, this error type is never used. ResLogic constants parsing related?
     ParseResLogicError,
     /// Occurs when values of different types are subtracted.
     TypeMismatchNotFelt,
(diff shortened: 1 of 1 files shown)


src/vm/error.zig
```diff
@@ -1,27 +1,52 @@
...
     InstructionFetchingFailed,
+    /// Error in converting the encoded instruction to a u64.
     InstructionEncodingError,
+    /// TODO, Logic error during parsing of the resolution logic?
     ParseResLogicError,
+    /// Occurs when values of different types are subtracted.
     TypeMismatchNotFelt,
+    /// Error encountered with a built-in runner.
     RunnerError,
+    /// Occurs when the expected value is not a Relocatable (e.g. not a Felt),
```
lana-shanghai, line 21: A `Relocatable` is never a `Felt`. A `MaybeRelocatable` can be a `Relocatable` or a `Felt`. So the e.g. example is not correct

src/vm/error.zig
```diff
@@ -1,27 +1,52 @@
...
+    /// or when subtracting two relocatables with different segment indices.
     TypeMismatchNotRelocatable,
+    /// Value is too large to be coerced to a u64.
     ValueTooLarge,
+    /// Occurs when both built-in deductions and fallback deductions for the operands fail.
     FailedToComputeOperands,
+    /// No destination register can be deduced for the given opcode.
     NoDst,
+    /// Occurs when both built-in deductions and fallback deductions fail to deduce Op1.
     FailedToComputeOp1,
```
lana-shanghai, line 31: Are we missing `FailedToComputeOp0`? :)

src/vm/error.zig
```diff
@@ -1,27 +1,52 @@
...
+    /// Multiplying with a relocatable is forbidden.
     MulRelocForbidden,
+    /// TODO, Invalid memory address encountered.
     InvalidMemoryAddress,
+    /// Failed to fetch instruction from VM during the instruction cycle.
     InstructionFetchingFailed,
+    /// Error in converting the encoded instruction to a u64.
     InstructionEncodingError,
+    /// TODO, Logic error during parsing of the resolution logic?
     ParseResLogicError,
```
lana-shanghai, line 16: Hmm I don't see this used anywhere. Do we need it?

src/vm/error.zig
lana-shanghai: I just checked and in tests for `computeOp0Deductions` we use the error `FailedToComputeOperands` we should probably add and use `FailedToComputeOp0 `

lana-shanghai (approved):
lgtm

PR #150, opened 2023-11-15
feat: add `AddressSet` structure for memory

### LINKED ISSUE #116: feat: add `AddressSet` structure for memory
At the moment, our validated addresses inside the `Memory` structure are contained inside a HashMap:

https://github.com/keep-starknet-strange/cairo-zig/blob/d677ee6aa56883d9ea6a26556be4e01a3404fa8e/src/vm/memory/memory.zig#L41-L46

To enhance this implementation, it could be better to wrap this inside a dedicated `AddressSet` data structure with dedicated methods as it is done in the Rust implementation [here](https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/vm/vm_memory/memory.rs#L44-L82). This will extend the specific functionality and properties of this data structure.

tcoratger:
Should close #116.

### DIFF
--- a/src/vm/memory/memory.zig
+++ b/src/vm/memory/memory.zig
@@ -45,13 +45,11 @@ pub const MemoryCell = struct {
     }
 };
 
-// ###############################################
-// ###############################################
-// ###############################################
-
+/// Represents a set of validated memory addresses in the Cairo VM.
 pub const AddressSet = struct {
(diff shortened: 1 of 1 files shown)


lana-shanghai (approved):
lgtm

PR #151, opened 2023-11-15
feat: add allow_tmp_segments flag in `computeEffectiveSizes`

### LINKED ISSUE #130: feat: add `allow_tmp_segments` flag in `computeEffectiveSizes`
The `computeEffectiveSize` function currently lacks a parameter to control the inclusion of temporary segments in the computation of effective sizes:

https://github.com/keep-starknet-strange/cairo-zig/blob/9af9d13cd5231fead222a23efe09fafa1a9c1d4f/src/vm/memory/segments.zig#L169-L180

This enhancement proposes the addition of a `allow_tmp_segments` boolean flag to the function's signature, enabling users to decide whether temporary segments should be considered.

**Enhancement Proposal:**
```
pub fn computeEffectiveSize(self: *Self, allow_tmp_segments: bool) !std.AutoArrayHashMap(u32, u32) {
    // Implementation with the new allow_tmp_segments flag.
    // ...
    // ...
}
```

**Additional Context:**
The addition of the `allow_tmp_segments` flag offers users more control over the computation of effective sizes, aligning the Zig implementation with the Python counterpart.

```
// Example Usage
const allowTmpSegments: bool = true; // Set to true to include temporary segments
const effectiveSizes = myCairoVM.computeEffectiveSize(allowTmpSegments);
```

**Reference Python implementation:** 
https://github.com/starkware-libs/cairo-lang/blob/10d257de5f49543757599cad9afaa5b4dd6d752b/src/starkware/cairo/lang/vm/memory_segments.py#L75-L96

itzlambda:
fix: #130

### DIFF
--- a/src/vm/memory/segments.zig
+++ b/src/vm/memory/segments.zig
@@ -182,16 +182,30 @@ pub const MemorySegmentManager = struct {
             first_segment_index = 0;
         }
 
-        while (first_segment_index < self.memory.num_segments) {
+        var first_index = first_segment_index;
+        while (first_index < self.memory.num_segments) {
             try self.segment_used_sizes.put(first_segment_index, 0);
-            first_segment_index += 1;
+            first_index += 1;
         }
 
-        for (self.memory.data.items, 0..) |segment, i| {
+        var temp1: i64 = 0;
+        for (self.memory.data.items) |segment| {
             try self.segment_used_sizes.put(
-                @intCast(i),
+                @intCast(temp1),
                 @intCast(segment.items.len),
             );
+            temp1 += 1;
+        }
+
+        if (allow_tmp_segments) {
+            var temp2: i64 = 0;
+            for (self.memory.temp_data.items) |segment| {
+                try self.segment_used_sizes.put(
+                    -@as(i64, temp2 + 1),
+                    @intCast(segment.items.len),
+                );
+                temp2 += 1;
+            }
         }
         return self.segment_used_sizes;
     }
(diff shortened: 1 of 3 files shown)


lambda-0x:
i didn't find any tests particularly for this in the reference implementation linked

src/vm/memory/segments.zig
```diff
@@ -167,7 +167,25 @@ pub const MemorySegmentManager = struct {
     /// # Returns
     ///
     /// An `AutoArrayHashMap` representing the computed effective sizes of memory segments.
-    pub fn computeEffectiveSize(self: *Self) !std.AutoArrayHashMap(u32, u32) {
+    pub fn computeEffectiveSize(self: *Self, allow_tmp_segments: bool) !std.AutoArrayHashMap(i64, u32) {
+        if (self.segment_used_sizes.count() != 0) {
+            return self.segment_used_sizes;
+        }
+
+        // TODO: Check if memory is frozen. At the time of writting this function memory cannot be frozen so we cannot check if it frozen.
```
lambda-0x, line 175: python implementation asserts that memory is frozen, do we have any equivalent for it?

tcoratger: No we don't, to be added in another PR

src/vm/memory/segments.zig
```diff
@@ -168,7 +168,25 @@ pub const MemorySegmentManager = struct {
     /// # Returns
     ///
     /// An `AutoArrayHashMap` representing the computed effective sizes of memory segments.
-    pub fn computeEffectiveSize(self: *Self) !std.AutoArrayHashMap(u32, u32) {
+    pub fn computeEffectiveSize(self: *Self, allow_tmp_segments: bool) !std.AutoArrayHashMap(i64, u32) {
```
tcoratger, line 171: Can we also add unit tests for `allow_tmp_segments = true` in order to validate this new feature.

lambda-0x: > try to check is the obtained size the all the segments are correct with the allow_tmp_segments flag set to true.

that is what i dont know, is there any good doc to understand how memory segments work?

tcoratger: You can look in the unit tests that are already done, for example this one: "MemorySegmentManager: computeEffectiveSize for one segment memory with gaps".

When you have:
```zig
try memoryFile.setUpMemory(
         memory_segment_manager.memory,
         std.testing.allocator,
         .{
             .{ .{ 0, 3 }, .{1} },
             .{ .{ 0, 4 }, .{1} },
             .{ .{ 0, 7 }, .{1} },
             .{ .{ 0, 9 }, .{1} },
         },
     );
```

For now you have:
```zig
var actual = try memory_segment_manager.computeEffectiveSize();
try expectEqual(@as(usize, 1), actual.count());
try expectEqual(@as(u32, 10), actual.get(0).?);
```

You can test the same thing by adding negative indexes by testing activating the `allow_tmp_segments ` flag.

src/vm/memory/segments.zig
```diff
@@ -168,12 +168,44 @@ pub const MemorySegmentManager = struct {
...
+        }
+
+        // TODO: Check if memory is frozen. At the time of writting this function memory cannot be frozen so we cannot check if it frozen.
+        var first_segment_index: i64 = undefined;
+
+        if (allow_tmp_segments) {
+            first_segment_index = -@as(i64, self.memory.num_temp_segments);
+        } else {
+            first_segment_index = 0;
+        }
```
tcoratger, line 183: I would use instead:

```zig
const first_segment_index: i64 = if (allow_tmp_segments) -@intCast(self.memory.num_temp_segments) else 0;
```

src/vm/memory/segments.zig
```diff
@@ -168,12 +168,44 @@ pub const MemorySegmentManager = struct {
...
+
+        var first_index = first_segment_index;
+        while (first_index < self.memory.num_segments) {
+            try self.segment_used_sizes.put(first_segment_index, 0);
+            first_index += 1;
+        }
```
tcoratger, line 189: For me you don't need this entire while stuff. I indeed see that it is done this way in the Python implementation but it is to initialize all sizes to 0 in case the segment contains nothing. For me the 0 will be automatically taken into account when you do `@intCast(segment.items.len)` below. In all cases this while is too expensive as it reviews all the segments while we do exactly the same loop afterwards.

src/vm/memory/segments.zig
```diff
@@ -168,12 +168,44 @@ pub const MemorySegmentManager = struct {
...
+
+        var temp1: i64 = 0;
+        for (self.memory.data.items) |segment| {
```
tcoratger, line 192: here there is no need for the `temp1` variable. The following is equivalent:

```zig
for (self.memory.data.items, 0..) |segment, i| {
...
}
```

src/vm/memory/segments.zig
```diff
@@ -168,12 +168,44 @@ pub const MemorySegmentManager = struct {
...
-                @intCast(i),
+                @intCast(temp1),
                 @intCast(segment.items.len),
             );
+            temp1 += 1;
+        }
+
+        if (allow_tmp_segments) {
+            var temp2: i64 = 0;
+            for (self.memory.temp_data.items) |segment| {
```
tcoratger, line 202: Here this is the same you can do directly:

```zig
for (self.memory.temp_data.items, 0..) |segment, i| {
```

src/vm/memory/segments.zig
lambda-0x: actually i tried doing that previously but for loop didn't support `i64`, but after the refactor don't need to loop over negatives values since they are separate arrays. so will update this.

src/vm/memory/segments.zig
lambda-0x: make sense, i was confused how it plays with the new structure so i left it there. will remove.

lambda-0x:
@tcoratger looks like there is a bug in `setupMemory` helper method, it doesn't properly setup `num_segments` and `num_temp_segments` variable in the memory

tcoratger:
> @tcoratger looks like there is a bug in `setupMemory` helper method, it doesn't properly setup `num_segments` and `num_temp_segments` variable in the memory

Should be handled by this one: https://github.com/keep-starknet-strange/cairo-zig/pull/169

tcoratger (approved):

PR #160, opened 2023-11-16
feat: add `add` and `sub` utils in `MaybeRelocatable` union

### LINKED ISSUE #158: feat: add `add` and `sub` utils in `MaybeRelocatable` union
At the moment, we don't have any util to add and subtract two `MaybeRelocatable` together. We need to implement this in the `MaybeRelocatable` structure:
- The `add` method performs addition between two MaybeRelocatable values. It checks the type of the operands and performs addition based on the types. Reference for the `add` implementation: https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/types/relocatable.rs#L235-L251
- The `sub` method, on the other hand, handles subtraction operations. Similar to `add`, it checks the types of the operands and performs subtraction accordingly. Reference for the `sub` implementation: https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/types/relocatable.rs#L253-L286

For Zig, we would need to replicate this logic, ensuring proper handling of different types of `MaybeRelocatable` values during addition and subtraction. Zig syntax and type system will differ from Rust, so you'll need to adapt this logic accordingly while maintaining the same behavior and rules for handling these types of operations.

tcoratger:
Should close #158.

### DIFF
--- a/src/vm/memory/relocatable.zig
+++ b/src/vm/memory/relocatable.zig
@@ -1,6 +1,7 @@
 const std = @import("std");
 const Felt252 = @import("../../math/fields/starknet.zig").Felt252;
 const CairoVMError = @import("../error.zig").CairoVMError;
+const MathError = @import("../error.zig").MathError;
 
 // Relocatable in the Cairo VM represents an address
 // in some memory segment. When the VM finishes running,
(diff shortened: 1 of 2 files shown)


src/vm/memory/relocatable.zig
```diff
@@ -887,6 +1018,62 @@ test "MaybeRelocatable: any comparision should return false if other MaybeReloca
...
+}
+
+test "MaybeRelocatable: add between a Felt252 and a Relocatable should return a proper MaybeRelocatable" {
+    try expectEqual(
+        fromSegment(0, 20),
+        try fromU256(10).add(fromSegment(0, 10)),
+    );
+}
+
+test "MaybeRelocatable: sub between two Relocatable should return a proper MaybeRelocatable" {
```
lana-shanghai, line 1049: extra test for `sub` between two Relocatables with different segment index?

tcoratger: @lana-shanghai Done :)

lana-shanghai (approved):
lgtm

PR #166, opened 2023-11-18
Fix/fib run

oxlime:
Think that perhaps the issue is not putting the deduced operands into memory?

### DIFF
--- a/src/vm/core.zig
+++ b/src/vm/core.zig
@@ -151,7 +151,11 @@ pub const CairoVM = struct {
         }
 
         const operands_result = try self.computeOperands(instruction);
-        _ = operands_result;
+
+        try self.updateRegisters(
+            instruction,
+            operands_result,
+        );
     }
 
     /// Compute the operands for a given instruction.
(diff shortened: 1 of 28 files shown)


src/vm/core.zig
```diff
@@ -256,6 +256,12 @@ pub const CairoVM = struct {
         return self.runInstruction(allocator, &instruction);
     }
 
+    pub fn insertDeducedOperands(self: *Self, allocator: Allocator, op: OperandsResult) !void {
+        try self.segments.memory.set(allocator, op.op_0_addr, op.op_0);
```
tcoratger, line 260: I think we must only do that if op0 was deducted: https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/vm/vm_core.rs#L369

Maybe we can implement the util logic https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/vm/vm_core.rs#L54-L74 in our `OperandsResult` structure wdt?

src/vm/core.zig
```diff
@@ -256,6 +256,12 @@ pub const CairoVM = struct {
...
+        try self.segments.memory.set(allocator, op.op_1_addr, op.op_1);
```
tcoratger, line 261: Same as before but with op1 no?

src/vm/core.zig
```diff
@@ -256,6 +256,12 @@ pub const CairoVM = struct {
...
+        try self.segments.memory.set(allocator, op.dst_addr, op.dst);
```
tcoratger, line 262: Same as before but with dest no?

src/vm/core.zig
```diff
@@ -256,6 +256,12 @@ pub const CairoVM = struct {
         return self.runInstruction(allocator, &instruction);
     }
 
+    pub fn insertDeducedOperands(self: *Self, allocator: Allocator, op: OperandsResult) !void {
```
tcoratger, line 259: Can we add unit tests to confirm the logic here?

src/vm/core.zig
oxlime: Are u thinking something like this 
https://github.com/oxlime/cairo-zig/blob/99173d5a81ddac28d6538d15609f0a59690432d8/src/vm/core.zig#L835

src/vm/core.zig
oxlime: Let me know if we need more tests

tcoratger (approved):

PR #169, opened 2023-11-19
Feat/set memory

### LINKED ISSUE #124: feat: complete `set` function of the `Memory` structure
At the moment the implementation of the `set` function inside the `Memory` structure is not complete:

https://github.com/keep-starknet-strange/cairo-zig/blob/f7065762676d169052e9e19b3049569107ec2428/src/vm/memory/memory.zig#L98-L125

There is a lack of checks and verifications to ensure, for example, that the added element is inserted next to the last one on the segment or that another element already exists at this address to possibly return an error.

Here are the implementations that can guide our Zig implementation:
- Implementation in Rust: https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/vm/vm_memory/memory.rs#L119-L165
- Python implementation: https://github.com/starkware-libs/cairo-lang/blob/10d257de5f49543757599cad9afaa5b4dd6d752b/src/starkware/cairo/lang/vm/memory_dict.py#L189

oxlime:
Closes #124 

more checks and tests needed

### DIFF
--- a/src/vm/builtins/builtin_runner/keccak.zig
+++ b/src/vm/builtins/builtin_runner/keccak.zig
@@ -343,9 +343,9 @@ pub const KeccakBuiltinRunner = struct {
             const stop_pointer_addr = pointer.subUint(
                 @intCast(1),
             ) catch return RunnerError.NoStopPointer;
-            const stop_pointer = try (segments.memory.get(
+            const stop_pointer = try ((segments.memory.get(
                 stop_pointer_addr,
-            ) catch return RunnerError.NoStopPointer).tryIntoRelocatable();
+            ) catch return RunnerError.NoStopPointer) orelse return RunnerError.NoStopPointer).tryIntoRelocatable();
             if (@as(
                 isize,
                 @intCast(self.base),
@@ -427,9 +427,9 @@ pub const KeccakBuiltinRunner = struct {
             usize,
             @intCast(self.n_input_cells),
         )) |i| {
-            const num = (memory.get(try first_input_addr.addUint(@intCast(i))) catch {
+            const num = ((memory.get(try first_input_addr.addUint(@intCast(i))) catch {
                 return null;
-            }).tryIntoFelt() catch {
+            }) orelse return null).tryIntoFelt() catch {
                 return RunnerError.BuiltinExpectedInteger;
             };
 
--- a/src/vm/builtins/builtin_runner/keccak.zig
+++ b/src/vm/builtins/builtin_runner/keccak.zig
@@ -837,6 +837,12 @@ test "KeccakBuiltinRunner: finalStack should return TypeMismatchNotRelocatable e
     var memory_segment_manager = try MemorySegmentManager.init(std.testing.allocator);
     defer memory_segment_manager.deinit();
 
+    try memory_segment_manager.memory.data.append(std.ArrayListUnmanaged(?MemoryCell){});
+    try memory_segment_manager.memory.data.append(std.ArrayListUnmanaged(?MemoryCell){});
+    try memory_segment_manager.memory.data.append(std.ArrayListUnmanaged(?MemoryCell){});
+    _ = memory_segment_manager.addSegment();
+    _ = memory_segment_manager.addSegment();
+    _ = memory_segment_manager.addSegment();
     try memory_segment_manager.memory.set(
         std.testing.allocator,
         try Relocatable.new(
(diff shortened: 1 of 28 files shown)


src/vm/memory/memory.zig
```diff
@@ -223,6 +223,10 @@ pub const Memory = struct {
         var data = if (address.segment_index < 0) &self.temp_data else &self.data;
         const segment_index: usize = @intCast(if (address.segment_index < 0) -(address.segment_index + 1) else address.segment_index);
 
+        if (self.num_segments <= segment_index) {
+            return MemoryError.UnallocatedSegment;
```
tcoratger, line 227: Can we add an unit test to check that?

src/vm/memory/memory.zig
```diff
@@ -239,6 +243,13 @@ pub const Memory = struct {
                 @as(usize, @intCast(address.offset)) + 1 - data_segment.items.len,
             );
         }
+
+        // check if existing memory, cannot overwrite
+        if (data_segment.items[@as(usize, @intCast(address.offset))] != null) {
+            if (!data_segment.items[@as(usize, @intCast(address.offset))].?.maybe_relocatable.eq(value)) {
+                return MemoryError.DuplicatedRelocation;
+            }
+        }
```
tcoratger, line 252: Here for me we can replace that by:

```zig
if (data_segment.items[@intCast(address.offset)]) |item| {
  if (!item.maybe_relocatable.eq(value)) {
      return MemoryError.DuplicatedRelocation;
  }
}
```

for a less verbose implementation :)

src/vm/builtins/builtin_runner/keccak.zig
```diff
@@ -837,6 +837,12 @@ test "KeccakBuiltinRunner: finalStack should return TypeMismatchNotRelocatable e
     var memory_segment_manager = try MemorySegmentManager.init(std.testing.allocator);
     defer memory_segment_manager.deinit();
 
+    try memory_segment_manager.memory.data.append(std.ArrayListUnmanaged(?MemoryCell){});
+    try memory_segment_manager.memory.data.append(std.ArrayListUnmanaged(?MemoryCell){});
+    try memory_segment_manager.memory.data.append(std.ArrayListUnmanaged(?MemoryCell){});
```
tcoratger, line 842: If I understand correctly, here you add empty `ArrayListUnmanaged` because you have to add 3 segments manually to have access to index 2.

If that's it, in my opinion, the logic for adding empty arrays must be found directly in the `addSegment` function: here it's a little different from Rust, when adding a segment, we can't just add 1 to the number of segments. We must also allocate an empty array there.

src/vm/core_test.zig
```diff
@@ -1263,8 +1254,8 @@ test "compute operands mul FP" {
     var vm = try CairoVM.init(allocator, .{});
     defer vm.deinit();
 
-    _ = vm.addMemorySegment();
-    _ = vm.addMemorySegment();
+    //    _ = vm.addMemorySegment();
+    //    _ = vm.addMemorySegment();
```
tcoratger, line 1258: We can delete that no?

src/vm/memory/memory.zig
```diff
@@ -470,14 +503,14 @@ test "Memory: validate existing memory" {
...
     );
-    try expectEqual(
-        false,
-        segments.memory.validated_addresses.contains(Relocatable.new(1, 1)),
-    );
-    try expectEqual(
-        false,
-        segments.memory.validated_addresses.contains(Relocatable.new(2, 2)),
-    );
+    //    try expectEqual(
```
tcoratger, line 506: Why has this changed?

src/vm/memory/memory.zig
```diff
@@ -654,6 +691,11 @@ test "memory set and get" {
     );
     const value_2 = relocatable.fromFelt(starknet_felt.Felt252.one());
 
+    try memory.data.append(std.ArrayListUnmanaged(?MemoryCell){});
+    memory.num_segments += 1;
+    try memory.temp_data.append(std.ArrayListUnmanaged(?MemoryCell){});
+    memory.num_segments += 1;
```
tcoratger, line 697: Here, you then use the `setUpMemory` function and it seems that you already take into account the addition of segments in this function, right? 

In this case there is no need to do it manually above, right? Which is preferable from a scalability point of view for the rest of our tests as well. We need to manage the addition of segments directly in the `setUpMemory` function to be able to setup our tests with as few manual procedures as possible.

src/vm/memory/memory.zig
```diff
@@ -419,6 +447,7 @@ pub fn setUpMemory(memory: *Memory, allocator: Allocator, comptime vals: anytype
             switch (@typeInfo(@TypeOf(row[1][0]))) {
                 .Pointer => {
                     try memory.set(
+                        //var res = memory.set(
```
tcoratger, line 450: Can you just remove this and for me we are good to go?

tcoratger (approved):

PR #171, opened 2023-11-19
refactor: use `setUpMemory` function to set up memory

### LINKED ISSUE #170: refactor: use `setUpMemory` function to set up memory
We still have some places in the code, especially for unit tests, where the `setUpMemory` function is not used to set up memory which is set up manually instead:

https://github.com/keep-starknet-strange/cairo-zig/blob/fd7ce5a27c37b15e7273ca2d8a187672f41a40d6/src/vm/memory/memory.zig#L483-L493

To facilitate the maintenance of the code and its readability, it would be preferable to refactor all these parts to use 100% of the time the `setUpMemory` function dedicated to this.

tcoratger:
Should close #170.

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -11,7 +11,7 @@
     .dependencies = .{
         .@"zig-cli" = .{
             .url = "https://github.com/sam701/zig-cli/archive/refs/heads/main.tar.gz",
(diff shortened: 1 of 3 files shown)


lana-shanghai (approved):
lgtm

PR #173, opened 2023-11-20
refactor: label relocatable errors

### LINKED ISSUE #161: refactor: better labels for errors in `Relocatable` and `MaybeRelocatable`
At the moment, in both the `Relocatable` and `MaybeRelocatable` structures, we have errors that are not labeled properly using our error types that can be found in the `error.zig` file.
- Example in `Relocatable` structure: https://github.com/keep-starknet-strange/cairo-zig/blob/d50e5492e15d454f7882703df4824866543aa5a2/src/vm/memory/relocatable.zig#L114
- Example in `MaybeRelocatable` union: https://github.com/keep-starknet-strange/cairo-zig/blob/d50e5492e15d454f7882703df4824866543aa5a2/src/vm/memory/relocatable.zig#L391

We need to label these errors properly in order to be able to group them and enforce a better typing.

dhruvkelawala:
Closes #161

### DIFF
--- a/src/vm/memory/relocatable.zig
+++ b/src/vm/memory/relocatable.zig
@@ -112,7 +112,7 @@ pub const Relocatable = struct {
     /// A new Relocatable after the subtraction operation.
     /// # Errors
     /// An error is returned if the subtraction results in an underflow (negative value).
-    pub fn subFelt(self: Self, other: Felt252) !Self {
+    pub fn subFelt(self: Self, other: Felt252) MathError!Self {
         return try self.subUint(try other.tryIntoU64());
     }
 
@@ -124,7 +124,7 @@ pub const Relocatable = struct {
     pub fn subUint(
         self: Self,
         other: u64,
-    ) !Self {
+    ) MathError!Self {
         if (self.offset < other) {
             return MathError.RelocatableSubUsizeNegOffset;
         }
@@ -196,7 +196,7 @@ pub const Relocatable = struct {
     /// A new Relocatable after the addition operation.
     /// # Errors
     /// An error is returned if the addition results in an overflow (exceeding u64).
-    pub fn addFelt(self: Self, other: Felt252) !Self {
+    pub fn addFelt(self: Self, other: Felt252) MathError!Self {
         return .{
             .segment_index = self.segment_index,
             .offset = try Felt252.fromInteger(@intCast(self.offset)).add(other).tryIntoU64(),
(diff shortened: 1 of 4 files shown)


src/vm/memory/relocatable.zig
```diff
@@ -420,7 +420,7 @@ pub const MaybeRelocatable = union(enum) {
     pub fn tryIntoRelocatable(self: Self) !Relocatable {
```
tcoratger, line 420: Here, as a result of your refactor, can we enforce a better typing of the output? like:

```zig
pub fn tryIntoRelocatable(self: Self) CairoVMError!Relocatable {
```

src/vm/memory/relocatable.zig
```diff
@@ -126,7 +126,7 @@ pub const Relocatable = struct {
         other: u64,
     ) !Self {
```
tcoratger, line 127: Here, as a result of your refactor, can we enforce a better typing of the output? like:

```zig
pub fn subUint(
        self: Self,
        other: u64,
    ) MathError!Self {
```

src/vm/memory/relocatable.zig
```diff
@@ -96,7 +96,7 @@ pub const Relocatable = struct {
     /// This method fails if `self` and other` are not from the same segment.
     pub fn sub(self: Self, other: Self) !Self {
```
tcoratger, line 97: Here, as a result of your refactor, can we enforce a better typing of the output?

src/vm/memory/relocatable.zig
```diff
@@ -777,7 +777,7 @@ test "Relocatable: addFelt should add a relocatable and a Felt252" {
 
 test "Relocatable: addFelt should return an error if number after offset addition is too large" {
     try expectError(
-        error.ValueTooLarge,
+        CairoVMError.ValueTooLarge,
```
tcoratger, line 780: Can we move this one to a `MathError`?

src/vm/memory/relocatable.zig
lana-shanghai: @tcoratger good point I will add a todo for that

lana-shanghai: ah there's more comments, OK let's wait for Dhruv to fix it

dhruvkelawala:
So I have added stricter type checking for errors in most of the functions. In the remaining ones, I think we need a better error handling approach and cannot be fixed just by type checker. For example, 
https://github.com/dhruvkelawala/cairo-zig/blob/408cf6fb246af92d3d1c13eac344510ee2db6278/src/vm/memory/relocatable.zig#L98

This function can return CairoVMError or MathError but idts that Zig supports return either one of those. 

Something like `error{CairoVMError, MathError}!Self`. We can handle each error in this function separately and return proper errors. This should be done in a separate PR imo

tcoratger (approved):

PR #177, opened 2023-11-21
zig: make new zig master build happy

tcoratger:
This PR contains:

- Migration from `var` to `const` when needed after recent update of zig version.
- Small change in `validateMemoryCell` due to `rangeCheckValidationRule` that is broken and needed to be better unit tested before being reactivated.


lana-shanghai (approved):
lgtm

PR #179, opened 2023-11-22
chore: add utilTestInstruction

### LINKED ISSUE #136: tests: add `Instruction` instance test util variable
## Overview

Currently, the `Instruction` structure is extensively used in various unit tests, such as in:

https://github.com/keep-starknet-strange/cairo-zig/blob/194de20850dcab37541385c7cce5d7773da67535/src/vm/core_test.zig#L1468-L1480

The current approach involves redeclaring a new instance of the `Instruction` structure for each test, resulting in cumbersome and verbose code.

## Proposed Enhancement

To improve the clarity and maintainability of the code, I recommend a different approach. Instead of frequently invoking the `Instruction.default()` function, which may lead to huge computational work (calling `decode` function each time) and potential issues if the function fails, I propose introducing a utility test variable. This variable will store a default structure, enabling easy reuse and modification across all unit tests.

## Benefits

1. **Readability**: The proposed approach streamlines the code, making it more readable by reducing redundancy.
2. **Maintainability**: A centralized test variable facilitates easy modifications across tests, enhancing maintainability.
3. **Error Isolation**: In the event of an issue with `Instruction.default()`, this approach isolates the problem to the utility variable, making troubleshooting more straightforward.

## Example Implementation

```zig
const utilTestInstruction: Instruction = Instruction.default();

// Example usage in a unit test
test "Some test case" {
    // Modify utilTestInstruction as needed for this specific test
    utilTestInstruction.someField = someValue;

    // Rest of the test logic
    // ...
}

dhruvkelawala:
Closes #136

### DIFF
--- a/src/vm/core_test.zig
+++ b/src/vm/core_test.zig
@@ -36,7 +36,7 @@ const expectError = std.testing.expectError;
 const expectEqualSlices = std.testing.expectEqualSlices;
 
 // Default Test Instruction to avoid having to initialize it in every test
-const utilTestInstruction = Instruction.default();
+const defaultTestInstruction = Instruction.default();
 
 test "CairoVM: deduceMemoryCell no builtin" {
     var vm = try CairoVM.init(
@@ -87,7 +87,7 @@ test "CairoVM: deduceMemoryCell builtin valid" {
 test "update pc regular no imm" {
     // Test setup
     const allocator = std.testing.allocator;
-    var instruction = utilTestInstruction;
+    var instruction = defaultTestInstruction;
     instruction.pc_update = .Regular;
     instruction.op_1_addr = .AP;
     const operands = OperandsResult.default();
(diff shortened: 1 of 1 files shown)


tcoratger:
For your information, there are other places where we have very verbose declarations that can be replaced:

https://github.com/keep-starknet-strange/cairo-zig/blob/fa08e433708b61f9878639fe5182c8afbfd2b1bf/src/vm/core_test.zig#L890-L902

dhruvkelawala:
Yes, I know. I didn't know what modifications would be required in the default instance to achieve the modified one. But I could try.

tcoratger:
> Yes, I know. I didn't know what modifications would be required in the default instance to achieve the modified one. But I could try.

Yes, if the default implementation does not work, we should try to find a pattern common to all declarations of this type and possibly create a separate variable that you can call `instruction_test` which groups most of the declarations. Indeed, if we look more closely, they are often similar and are then modified so I think we can proceed in this way to avoid redeclaring the same thing each time at the start of the test.

tcoratger (approved):

PR #180, opened 2023-11-22
feat: add memory comparison methods in `Memory`

### LINKED ISSUE #147: feat: add memory comparison methods in `Memory`
## Description
I would like to propose the implementation of `memcmp` and `mem_eq` methods in Ziglang's `Memory` structure, similar to the ones present in the [Rust](https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/vm/vm_memory/memory.rs#L349-L434). These methods will be used for comparing two ranges of values in memory.

## Proposed Changes
Add the following methods to the `Memory` structure in Ziglang:

```zig
// Compares two ranges of values in memory of length `len`
// Returns the ordering and the first relative position at which they differ
// Special cases:
// - `lhs` exists in memory but `rhs` doesn't -> (Ordering::Greater, 0)
// - `rhs` exists in memory but `lhs` doesn't -> (Ordering::Less, 0)
// - None of `lhs` or `rhs` exist in memory -> (Ordering::Equal, 0)
// Everything else behaves much like `memcmp` in C.
// This is meant as an optimization for hints to avoid allocations.
pub fn memCmp(lhs: Relocatable, rhs: Relocatable, len: usize) -> (Ordering, usize) {
    // Implementation here...
}

// Compares two ranges of values in memory of length `len`
// Returns true if the two ranges are equal, false otherwise
// Special cases:
// - `lhs` exists in memory but `rhs` doesn't -> false
// - `rhs` exists in memory but `lhs` doesn't -> false
// - None of `lhs` or `rhs` exist in memory -> true
pub fn memEq(lhs: Relocatable, rhs: Relocatable, len: usize) -> bool {
    // Implementation here...
}
```

## Context
The proposed methods are used for comparing memory ranges efficiently, providing optimizations for hints to avoid unnecessary allocations. The Rust implementation provided in the reference file can serve as a guide for the Zig implementation.

tcoratger:
Should close #147.

### DIFF
--- a/src/vm/memory/memory.zig
+++ b/src/vm/memory/memory.zig
@@ -122,14 +122,20 @@ pub const MemoryCell = struct {
     ///
     /// Returns a `std.math.Order` representing the order relationship between
     /// the two MemoryCell instances.
-    pub fn cmp(self: Self, other: Self) std.math.Order {
-        return switch (self.maybe_relocatable.cmp(other.maybe_relocatable)) {
-            .eq => switch (self.is_accessed) {
-                true => if (other.is_accessed) .eq else .gt,
-                false => if (other.is_accessed) .lt else .eq,
-            },
-            else => |res| res,
-        };
+    pub fn cmp(self: ?Self, other: ?Self) std.math.Order {
+        if (self) |lhs| {
+            if (other) |rhs| {
+                return switch (lhs.maybe_relocatable.cmp(rhs.maybe_relocatable)) {
+                    .eq => switch (lhs.is_accessed) {
+                        true => if (rhs.is_accessed) .eq else .gt,
+                        false => if (rhs.is_accessed) .lt else .eq,
+                    },
+                    else => |res| res,
+                };
+            }
+            return .gt;
+        }
+        return if (other == null) .eq else .lt;
     }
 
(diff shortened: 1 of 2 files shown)


src/vm/memory/memory.zig
```diff
@@ -51,6 +52,139 @@ pub const MemoryCell = struct {
...
+    /// # Arguments
+    ///
+    /// - `self`: The first MemoryCell instance to compare.
+    /// - `other`: The second MemoryCell instance to compare against.
+    ///
+    /// # Returns
+    ///
+    /// Returns a `std.math.Order` representing the order relationship between
+    /// the two MemoryCell instances.
+    pub fn cmp(self: ?Self, other: ?Self) std.math.Order {
```
lana-shanghai, line 125: Why do the inputs need to be optional values?

tcoratger: Because our vectors which contain the memory elements contain both `MemoryCell` but also `null` when there is nothing. However when we use this function for a `MemoryCell`, we do not know if it contains a `MaybeRelocatable` or a `null` which is why we must use an optional type.

lana-shanghai (approved):
lgtm

PR #187, opened 2023-11-23
feat: implementation of `getRange` method in `Memory` structure

### LINKED ISSUE #184: feat: implementation of `getRange` method in `Memory` structure
## Description

The Memory structure requires the implementation of a `getRange` method that retrieves a range of memory values from a specified starting address to a designated size. This method constructs a collection of memory values, allowing for potential gaps in cases where the original memory contains such discontinuities.

## Objective

The primary objective behind integrating the `getRange` method is to facilitate the retrieval of a sequence of memory values within the `Memory` structure. The method should accommodate potential gaps in memory, providing a comprehensive range of values within the specified address and size.

## Reference

Reference for the Rust implementation: https://github.com/lambdaclass/cairo-vm/blob/397e69910076d2c671b6b84618b20a34f6bdec56/vm/src/vm/vm_memory/memory.rs#L438-L446

tcoratger:
Should close #184.

### DIFF
--- a/src/vm/memory/memory.zig
+++ b/src/vm/memory/memory.zig
@@ -1001,9 +1001,9 @@ test "Memory: getRange for continuous memory" {
     var expected_vec = std.ArrayList(?MaybeRelocatable).init(std.testing.allocator);
     defer expected_vec.deinit();
 
-    try expected_vec.append(relocatable.fromU256(2));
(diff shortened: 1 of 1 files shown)


lana-shanghai (approved):
fix CI, then LGTM

PR #188, opened 2023-11-23
feat: implementation of `getContinuousRange` method in `Memory` structure

### LINKED ISSUE #185: feat: implementation of `getContinuousRange` method in `Memory` structure
## Description

The `Memory` structure necessitates the implementation of a `getContinuousRange` method that retrieves a contiguous range of memory values from a specified starting address to a designated size. This method ensures that no memory gaps exist within the retrieved sequence, failing if any value within the range is missing.

## Objective

The primary objective behind integrating the get_continuous_range method is to ensure the retrieval of a contiguous sequence of memory values within the Memory structure. Error handling is crucial to identify and handle scenarios where memory gaps exist within the specified range.

## Reference

Reference to the Rust implementation: https://github.com/lambdaclass/cairo-vm/blob/397e69910076d2c671b6b84618b20a34f6bdec56/vm/src/vm/vm_memory/memory.rs#L450-L465

tcoratger:
Should close #185.

### DIFF
--- a/src/vm/error.zig
+++ b/src/vm/error.zig
@@ -45,6 +45,7 @@ pub const MemoryError = error{
     TemporarySegmentInRelocation,
     /// Inconsistent Relocation
     Relocation,
+    GetRangeMemoryGap,
 };
 
 /// Reepresents different error conditions that occur in the built-in runners.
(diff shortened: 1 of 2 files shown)


lana-shanghai (approved):
lgtm

PR #198, opened 2023-11-25
feat: enhance `OperandsResult` structure and methods

tcoratger:
This PR focuses on:
- Enhancing the current `OperandsResult` structure with a better documentation for each field and each method.
- Using `@intFromBool` type coercion instead of `if` statements to transform boolean value to unsigned integers.

### DIFF
--- a/src/vm/core.zig
+++ b/src/vm/core.zig
@@ -846,16 +846,28 @@ pub fn deduceOp1(
 pub const OperandsResult = struct {
     const Self = @This();
 
+    /// The destination operand value.
     dst: MaybeRelocatable,
+    /// The result operand value.
     res: ?MaybeRelocatable,
+    /// The first operand value.
     op_0: MaybeRelocatable,
+    /// The second operand value.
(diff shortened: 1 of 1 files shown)


src/vm/core.zig
```diff
@@ -869,24 +881,56 @@ pub const OperandsResult = struct {
...
     }
 
+    /// Sets the flag indicating the second operand was deduced.
+    ///
+    /// # Arguments
+    ///
+    /// - `value`: A boolean value indicating whether the second operand was deduced.
     pub fn setOp1(self: *Self, value: bool) void {
-        self.deduced_operands |= if (value) 1 << 2 else 0 << 2;
+        self.deduced_operands |= @as(u8, @intCast(@intFromBool(value))) << 2;
```
lana-shanghai, line 908: why not something like `self.deduced_operands |= if (value) @as(u8, 1) << 2 else @as(u8, 0) << 2;` ?

lana-shanghai (approved):
Remove redundant coercion, then LGTM

PR #201, opened 2023-11-26
feat: improve `Relocatable` operation type coercion and error handling

### LINKED ISSUE #200: feat: add `addUsize` method in `Relocatable` structure
Throughout our codebase, there's a recurring need to add `usize` values to instances of the `Relocatable` structure. Presently, the absence of a dedicated method within the structure necessitates verbose integer casting operations, resulting in code verbosity and decreased readability:

https://github.com/keep-starknet-strange/cairo-zig/blob/afef208f58356f3b492c189235f8f1632fb5be76/src/vm/builtins/builtin_runner/keccak.zig#L432

To streamline and simplify these operations, the proposal is to introduce an `addUsize` method within the `Relocatable` structure. This method will facilitate direct addition of `usize` values to instances of `Relocatable`, promoting code clarity, reducing verbosity, and enhancing maintainability across the codebase.

tcoratger:
This PR should close #200.

Finally, there is no need to implement an `addUsize` method because it is implicitly taken into account by the coercion type during a call function described here: https://ziglang.org/documentation/master/#toc-Type-Coercion

Thus, this be satisfied with a call `addUint(x)` with `x` of type `usize` and the coercion of the type `usize` to `u64` is done implicitly without any problem because it is completely unambiguous how to get from one type to another.

This PR therefore focuses on the following points:
- Call the `addUint` function of the `Relocatable` structure without needing verbose `intCast` for unambiguous types like usize.
- Better error handling is also done for adding `u64` to an instance of `Relocatable` in the case of an overflow.
- A bit of refactoring for to transform comment to comment doc in `Relocatable` structure.

### DIFF
--- a/src/vm/builtins/bitwise/bitwise.zig
+++ b/src/vm/builtins/bitwise/bitwise.zig
@@ -60,7 +60,7 @@ fn getValue(address: Relocatable, memory: *Memory) BitwiseError!u256 {
 /// - memory: The cairo memory where addresses are looked up
 /// # Returns
 /// The deduced value as a `MaybeRelocatable`
-pub fn deduce(address: Relocatable, memory: *Memory) BitwiseError!MaybeRelocatable {
+pub fn deduce(address: Relocatable, memory: *Memory) !MaybeRelocatable {
     const index = address.offset % CELLS_PER_BITWISE;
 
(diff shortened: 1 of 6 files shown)


lana-shanghai (approved):
lgtm

PR #206, opened 2023-12-01
normalize Bitwise to BuiltinRunner conventions

jobez:
Conforms to the methods implemented in the bitwise builitin in rust. 

As there was functional overlap of methods in keccak, most added methods are reusing the logic from there, aside from:

- different `getAllocatedMemoryUnit` logic
- different `getUsedDilutedCheckUnits` logic 
- different deduceMemoryCell logic (obv hehe)

so test coverage is following keccak builtin, with extra unique tests to cover the implementation diffs


src/vm/builtins/builtin_runner/bitwise.zig
```diff
@@ -211,7 +231,7 @@ pub const BitwiseBuiltinRunner = struct {
             else => return BitwiseError.InvalidBitwiseIndex,
         };
 
-        return MaybeRelocatable{ .felt = Felt252.fromInteger(res) };
+        return .{ .felt = Felt252.fromInteger(res) };
```
tcoratger, line 234: Maybe here we can directly use: 

```zig
MaybeRelocatable.fromU256( switch (index) {
            2 => x & y, // and
            3 => x ^ y, // xor
            4 => x | y, // or
            else => return BitwiseError.InvalidBitwiseIndex,
        }
)
```

method no?

jobez: cool idea!

src/vm/builtins/builtin_runner/bitwise.zig
```diff
@@ -53,14 +93,516 @@ pub const BitwiseBuiltinRunner = struct {
...
+
+        while (i < total_n_bits) : (i += diluted_spacing * diluted_n_bits) {
+            for (0..diluted_spacing) |j| {
+                if (i + j < total_n_bits) {
+                    try partition.append(i + j);
+                }
+            }
+        }
+
+        const partition_length = @as(usize, partition.items.len);
```
tcoratger, line 298: I think that `len` field is already a `usize` natively no? hence maybe no need for `@as(usize,...)` coercion no?

src/vm/builtins/builtin_runner/bitwise.zig
```diff
@@ -53,14 +93,516 @@ pub const BitwiseBuiltinRunner = struct {
...
+    /// A `Relocatable` pointer to the final stack pointer, or an error code if the
+    /// verification fails.
+    pub fn finalStack(
+        self: *Self,
+        segments: *MemorySegmentManager,
+        pointer: Relocatable,
+    ) !Relocatable {
+        if (self.included) {
+            const stop_pointer_addr = pointer.subUint(
+                @intCast(1),
```
tcoratger, line 330: Not sure you need int casting here, 1 should be enough no?

src/vm/builtins/builtin_runner/bitwise.zig
```diff
@@ -24,6 +44,26 @@ pub const BitwiseBuiltinRunner = struct {
...
     instances_per_component: u32,
 
+    /// Retrieve the felt in memory that an address denotes as an integer within the configured `total_n_bit` limit,
+    /// default is 251
+    /// # Arguments
+    /// - address: The address belonging to the Bitwise builtin's segment
+    /// - memory: The cairo memory where addresses are looked up
+    /// # Returns
+    /// The felt as an integer.
+    fn getIntWithinBits(self: Self, address: Relocatable, memory: *Memory) BitwiseError!u256 {
```
tcoratger, line 54: why not using a pointer as first arg? like 

```zig
fn getIntWithinBits(self: *Self, address: Relocatable, memory: *Memory)
```

src/vm/builtins/builtin_runner/bitwise.zig
```diff
@@ -53,14 +93,516 @@ pub const BitwiseBuiltinRunner = struct {
         };
     }
 
+    pub fn initDefault() Self {
+        var default: bitwise_instance_def.BitwiseInstanceDef = .{};
+        return Self.init(&default, true);
```
tcoratger, line 98: Maybe in the `init` function we can set `instance_def` as a pointer to constant because this is not modified and this will permit you to do

```zig
return Self.init(&bitwise_instance_def.BitwiseInstanceDef = .{}, true);
```

So I think we need to modify `init` declaration to:
```zig
pub fn init(
        instance_def: *const bitwise_instance_def.BitwiseInstanceDef,
        included: bool,
    ) Self {
```

src/vm/builtins/builtin_runner/bitwise.zig
jobez: you'd have a pointer to self when you want to mutate state, right? lmk if im understanding

src/vm/builtins/builtin_runner/bitwise.zig
jobez: trying to use an assignment in passing an argument is giving me an error (feel free to sanity check me) but this works!
```
&@as(bitwise_instance_def.BitwiseInstanceDef, .{})
```

src/vm/types/bitwise_instance_def.zig
```diff
@@ -3,16 +3,18 @@ pub const CELLS_PER_BITWISE: u32 = 5;
...
 
     /// Ratio
-    ratio: ?u32,
+    ratio: ?u32 = 256,
     /// The number of bits in a single field element that are supported by the bitwise builtin.
-    total_n_bits: u32,
+    total_n_bits: u32 = TOTAL_N_BITS_BITWISE_DEFAULT,
 
-    pub fn init() Self {
+    pub fn initDefault() Self {
```
jobez, line 17: per discussion, do we want to have conventions for structs where we have 'initDefault' iff there's a constructor involved?

src/vm/builtins/builtin_runner/bitwise.zig
```diff
@@ -53,14 +75,1115 @@ pub const BitwiseBuiltinRunner = struct {
...
+    /// If the instance is marked as included, a single element initialized with the base address is returned.
+    pub fn initialStack(self: *Self, allocator: Allocator) !ArrayList(MaybeRelocatable) {
+        var result = ArrayList(MaybeRelocatable).init(allocator);
+        errdefer result.deinit();
+        if (self.included) {
+            try result.append(MaybeRelocatable.fromSegment(
+                @intCast(self.base),
+                0,
+            ));
+            return result;
```
tcoratger, line 119: As we always return `result` we can remove this one no? And get only the `return result;` statement outside of the if condition

src/vm/types/bitwise_instance_def.zig
tcoratger: Yes following our discussion of yesterday it seems that we can avoid this one and push default values directly in the structure def.

tcoratger (approved):

tcoratger (approved):

PR #207, opened 2023-12-02
fix: modify `get` method in `Memory` to return `null` instead of error

### LINKED ISSUE #205: bug: fibonacci evaluation errors on first step on `dst_addr` lookup
See https://github.com/keep-starknet-strange/cairo-zig/actions/runs/7055855307/job/19206936981?pr=157 for the failing case. Locally I compared our evaluation (right side) to the cairo-vm_in_go (left side) at commit e806055 (closer to the fibonacci milestone) and was seeing this difference:

telegram-cloud-photo-size-1-5132342471563520857-y

tcoratger:
Should close #205.


lana-shanghai (approved):
lgtm

PR #222, opened 2023-12-04
felt: implement `getFeltRange` in `Memory`

### LINKED ISSUE #193: feat: add `getFelt252Range` method in `Memory` structure
## Description

The `getFelt252Range` function is an essential method in Rust for retrieving a range of `Felt252` values from a given memory address over a specified size. This method ensures that the values within this range are not missing and are all of type `Felt252`.

## Objective

Request the equivalent implementation of the `getFelt252Range` function from Rust in Zig to ensure consistency in functionalities between languages and facilitate code portability or migration utilizing this feature.

## Example of Zig implementation

```zig
pub fn getFelt252Range(
    self: *Self,
    addr: Relocatable,
    size: usize,
) ![]const Felt252 {
    // Implementation of the function...
}
```

## Reference

Reference for the Rust implementation: https://github.com/lambdaclass/cairo-vm/blob/4089b469ae855414b098abed1fd15f73b03ab7e1/vm/src/vm/vm_memory/memory.rs#L470-L482

ptisserand:
This PR should close #193 

Please note I have to call `get` in case of `ExpectedInteger` to check for gap in range to avoid modification in `getFelt`.

### DIFF
--- a/src/vm/builtins/builtin_runner/range_check.zig
+++ b/src/vm/builtins/builtin_runner/range_check.zig
@@ -287,7 +287,7 @@ pub const RangeCheckBuiltinRunner = struct {
 /// verification fails.
 pub fn rangeCheckValidationRule(memory: *Memory, address: Relocatable) MemoryError![]const Relocatable {
     const num = memory.getFelt(address) catch |err| switch (err) {
-        error.MemoryOutOfBounds => return MemoryError.RangeCheckGetError,
+        error.UnknownMemoryCell => return MemoryError.RangeCheckGetError,
         error.ExpectedInteger => return MemoryError.RangecheckNonInt,
     };
 
--- a/src/vm/core_test.zig
+++ b/src/vm/core_test.zig
@@ -1639,7 +1639,7 @@ test "CairoVM: getSegmentSize should return the size of the segment via getSegme
     try expectEqual(@as(u32, 6), vm.getSegmentSize(3).?);
 }
 
-test "CairoVM: getFelt should return MemoryOutOfBounds error if no value at the given address" {
+test "CairoVM: getFelt should return UnknownMemoryCell error if no value at the given address" {
     // Test setup
     var vm = try CairoVM.init(
         std.testing.allocator,
@@ -1649,7 +1649,7 @@ test "CairoVM: getFelt should return MemoryOutOfBounds error if no value at the 
 
     // Test checks
     try expectError(
-        error.MemoryOutOfBounds,
+        error.UnknownMemoryCell,
         vm.getFelt(Relocatable.new(10, 30)),
     );
 }
--- a/src/vm/memory/memory.zig
+++ b/src/vm/memory/memory.zig
@@ -756,6 +756,46 @@ pub const Memory = struct {
         }
         return values;
     }
+
+    /// Retrieves a continuous range of `Felt252` values starting from a specified address.
+    ///
+    /// # Arguments
+    ///
+    /// * `address`: The starting address in the memory from which the continuous range of `Felt252` is retrieved.
+    /// * `size`: The size of the continuous range of `Felt252` to be retrieved.
+    ///
+    /// # Returns
+    ///
+    /// Returns a list containing `Felt252` values retrieved from the continuous range starting at the given address.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if there are any gaps encountered within the continuous memory range.
+    /// Returns an error if value inside the range is not a `Felt252`
+    pub fn getFeltRange(
+        self: *Self,
+        address: Relocatable,
+        size: usize,
+    ) !std.ArrayList(Felt252) {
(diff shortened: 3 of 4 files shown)


src/vm/memory/memory.zig
```diff
@@ -762,6 +762,46 @@ pub const Memory = struct {
...
+            const elem = self.getFelt(try address.addUint(@intCast(i))) catch |err| {
+                // getFelt return ExpectedInteger
+                if (err == error.ExpectedInteger) {
+                    if (null == try (self.get(try address.addUint(@intCast(i))))) {
+                        return MemoryError.GetRangeMemoryGap;
+                    }
+                }
+                return err;
+            };
+            try values.append(elem);
```
tcoratger, line 801: Here why do we need specific error modifications? I think that doing:

```zig
try values.append(try self.getFelt(try address.addUint(@intCast(i))));
```

is enough because we must not modify the errors returned by the `getFelt` method. I just saw that the errors in the `getFelt` method are not exactly correct, it is better that you modify them directly in the `getFelt` method itself which should look like:

```zig
pub fn getFelt(
         self: *Self,
         address: Relocatable,
     ) error{ MemoryOutOfBounds, ExpectedInteger }!Felt252 {
         if (try self.get(address)) |m| {
             return switch(m) {
                 .felt => |fe| fe,
                 else => MemoryError.ExpectedInteger,
             };
         } else {
             return MemoryError.UnknownMemoryCell;
         }
     }
```

You should then add these errors in `MemoryError`:

https://github.com/keep-starknet-strange/cairo-zig/blob/18dccfa07e7f1211c821b4502121af5cfd1e0e16/src/vm/error.zig#L37-L56

and maybe adjust some unit tests.

ptisserand: I have added `ExpectedInteger` and `UnknownMemoryCell` in `MemoryError` but I have to remove error set in `getFelt` since I didn't find a way to do a declaration like:
```zig
error { MemoryOutBounds} || MemoryError { ExpectedInteger, UnknownMemoryCell} ! Felt252
```

Test cases for `getFelt` have been updated.

tcoratger: Yes no problem fine :) This seems good to me, pending @lana-shanghai green light for merging and conflict resolution with main branch.

ptisserand: I have looked at conflict and I'm wondering if I should keep `MemoryError.ExpectedInteger` and `MemoryError.UnknowMemoryCell` error for `getFelt` or update `getFeltRange` to use `error.ExpectedInteger` and `error.MemoryOutOfBounds`

@tcoratger what do you think ?

tcoratger: @ptisserand For me we should stay with:

```zig
pub fn getFelt(
         self: *Self,
         address: Relocatable,
     ) error{ MemoryOutOfBounds, ExpectedInteger }!Felt252 {
         if (try self.get(address)) |m| {
             return switch(m) {
                 .felt => |fe| fe,
                 else => MemoryError.ExpectedInteger,
             };
         } else {
             return MemoryError.UnknownMemoryCell;
         }
     }
```

and not modify this error in `getFeltRange` which should inherit from the errors of `getFelt`

ptisserand: Ok, I have fixed conflict and rebased on main.

src/vm/builtins/builtin_runner/range_check.zig
```diff
@@ -287,7 +287,7 @@ pub const RangeCheckBuiltinRunner = struct {
 /// verification fails.
 pub fn rangeCheckValidationRule(memory: *Memory, address: Relocatable) MemoryError![]const Relocatable {
     const num = memory.getFelt(address) catch |err| switch (err) {
-        error.MemoryOutOfBounds => return MemoryError.RangeCheckGetError,
+        error.UnknownMemoryCell => return MemoryError.RangeCheckGetError,
```
tcoratger, line 290: Here we should have `MemoryError.UnknownMemoryCell`

src/vm/builtins/builtin_runner/range_check.zig
```diff
@@ -287,7 +287,7 @@ pub const RangeCheckBuiltinRunner = struct {
...
         error.ExpectedInteger => return MemoryError.RangecheckNonInt,
```
tcoratger, line 291: Here we should have `MemoryError.ExpectedInteger`

src/vm/core_test.zig
```diff
@@ -1649,7 +1649,7 @@ test "CairoVM: getFelt should return MemoryOutOfBounds error if no value at the
 
     // Test checks
     try expectError(
-        error.MemoryOutOfBounds,
+        error.UnknownMemoryCell,
```
tcoratger, line 1652: Here we should have `MemoryError.UnknownMemoryCell`

tcoratger (approved):

PR #251, opened 2023-12-11
feat: add output builtin implementation

### LINKED ISSUE #250: feat: add output builtin implementation
## Description
Implement the Output Builtin functionality in Ziglang. 

## Reference

- Reference to the Rust implementation: https://github.com/lambdaclass/cairo-vm/blob/main/vm/src/vm/runners/builtin_runner/output.rs

tcoratger:
Should close #250.


src/vm/builtins/builtin_runner/output.zig
```diff
@@ -24,14 +92,228 @@ pub const OutputBuiltinRunner = struct {
...
+        page_size: usize,
+    ) !void {
+        _ = page_size;
+        if (self.pages.get(page_id) != null) return OutputBuiltinRunnerError.PageIdAlreadyAssigned;
+
+        if (!page_start.isRelocatable() or page_start.relocatable.segment_index != self.base) {
+            return OutputBuiltinRunnerError.PageStartNotInOutputSegment;
+        }
+
+        // TODO: implement rest of the logic by modifying base to Relocatable
```
jobez, line 314: are you sure we need to change base to Relocatable?

assuming its always the case that base is initialized by having its own segment (i.e. a builtin will always have its own segment), isnt the segment idx always going to just be zero?

https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/cairo/lang/vm/output_builtin_runner.py#L27
https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/cairo/lang/vm/memory_segments.py#L42

lmk if theres a case im not seeing 👂️️️️

tcoratger: I'm not sure because here: https://github.com/starkware-libs/cairo-lang/blob/ee7ce74e1159a349d4b77a5f952241b50b1692de/src/starkware/cairo/lang/vm/output_builtin_runner.py#L27

When you initialize the segments, you give the base the value of the new segment. In the Rust implementation they chose to take only the index segment, hence their use of a `usize` instead of a `Relocatable`.

But then for example when you do the `add_page` operation, then on the one hand you have to make sure that the `page_start` and the `base` have the same indexed segment (https://github.com/starkware-libs/cairo-lang/blob/ee7ce74e1159a349d4b77a5f952241b50b1692de/src/starkware/cairo/lang/vm/output_builtin_runner.py#L106) but then you also have to do a global subtraction of the offsets here: https://github.com/starkware-libs/cairo-lang/blob/ee7ce74e1159a349d4b77a5f952241b50b1692de/src/starkware/cairo/lang/vm/output_builtin_runner.py#L108.

So in all cases for me there is an inconsistency:
- The `base` could be considered as the segment index only arising from the initialization of the segments as is done in Rust.
- But on the other hand we also need the offset for certain operations.

Did I miss something here?

jobez: Basically what I am asking is if theres ever a case that the base offset will not be zero. See how the rust impl constructs a relocatable from its base: 

https://github.com/lambdaclass/cairo-vm/blob/8a2ef24c1ca258f1d1d230cbbd1478cd1d4cf9db/vm/src/vm/runners/builtin_runner/range_check.rs#L74

So you could potentially do the same thing for the subtraction of relocatables that is happening in `addPage`

tcoratger: @jobez Yes you are right, implementation is done ready for review and merge.

It is only missing the `attributes` field which is used for hints in the Python implementation but which is also not present in the Rust implementation. We can implement it on another PR because it is not absolutely necessary at the moment.

src/vm/builtins/builtin_runner/output.zig
```diff
@@ -24,14 +92,271 @@ pub const OutputBuiltinRunner = struct {
...
+    ///
+    /// # Arguments
+    ///
+    /// - `self`: A pointer to the OutputBuiltinRunner instance.
+    /// - `segments`: A pointer to the MemorySegmentManager managing memory segments.
+    ///
+    /// # Returns
+    ///
+    /// An error if the addition of the segment fails, otherwise sets the base address successfully.
+    pub fn initializeSegments(self: *Self, segments: *MemorySegmentManager) !void {
```
jobez, line 116: what do you think about mapping uses of the word `initialize` in implementations to `init` in zig? I searched through tigerbeetle and the zig code base and that seems to be the convention. 

so here we would do  `initializeSegments` -> `initSegements`

src/vm/builtins/builtin_runner/output.zig
```diff
@@ -12,6 +58,28 @@ pub const OutputBuiltinRunner = struct {
...
+    /// Initializes a new instance of the OutputBuiltinRunner.
+    ///
+    /// This function creates and returns a new OutputBuiltinRunner instance
+    /// with default configuration, setting the base address to 0, stop pointer to null,
+    /// and marks the runner as included by default.
+    ///
+    /// # Returns
+    ///
+    /// A new instance of OutputBuiltinRunner with default settings.
+    pub fn init(allocator: Allocator) Self {
```
jobez, line 75: can this be `initDefault`? zig precedence for this:

https://github.com/ziglang/zig/blob/master/lib/std/enums.zig#L737

src/vm/builtins/builtin_runner/output.zig
```diff
@@ -24,14 +92,271 @@ pub const OutputBuiltinRunner = struct {
     /// # Returns
     ///
     /// A new `OutputBuiltinRunner` instance.
-    pub fn init(included: bool) Self {
+    pub fn from(included: bool, allocator: Allocator) Self {
```
jobez, line 95: in turn to the `initDefault` suggestion, can this be just `init`?

jobez (commented):
some comments just around conventions around naming. have pulled this locally to compare across the python and rust implementation for a deeper review

jobez (changes requested):

src/vm/builtins/builtin_runner/output.zig
tcoratger: Yep agreed, I just hadn't made the change afterwards because it was the same in the builtin keccak but I just changed everything like that it's done.

src/vm/builtins/builtin_runner/output.zig
tcoratger: Agreed done

src/vm/builtins/builtin_runner/output.zig
tcoratger: Agreed, done

src/vm/builtins/builtin_runner/output.zig
```diff
@@ -24,14 +92,271 @@ pub const OutputBuiltinRunner = struct {
...
+    ///
+    /// - `self`: A pointer to the OutputBuiltinRunner instance.
+    /// - `allocator`: The allocator to initialize the ArrayList.
+    ///
+    /// # Returns
+    ///
+    /// An ArrayList of MaybeRelocatable elements representing the initial stack.
+    /// If the instance is marked as included, a single element initialized with the base address is returned.
+    pub fn initialStack(self: *Self, allocator: Allocator) !ArrayList(MaybeRelocatable) {
+        var result = ArrayList(MaybeRelocatable).init(allocator);
```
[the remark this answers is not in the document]
tcoratger: Nice remark, done

src/vm/builtins/builtin_runner/output.zig
```diff
@@ -24,14 +92,271 @@ pub const OutputBuiltinRunner = struct {
...
+    ///
+    /// - `self`: A pointer to the OutputBuiltinRunner instance.
+    /// - `segments`: A pointer to the MemorySegmentManager managing memory segments.
+    ///
+    /// # Returns
+    ///
+    /// An error if the addition of the segment fails, otherwise sets the base address successfully.
+    pub fn initializeSegments(self: *Self, segments: *MemorySegmentManager) !void {
+        self.base = @intCast((try segments.addSegment()).segment_index);
+        self.stop_ptr = null;
```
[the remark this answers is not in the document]
tcoratger: You are right, removed.

jobez (approved):

PR #265, opened 2023-12-16
feat: Implement deserialization between `ProgramJson` and `Program`

### LINKED ISSUE #256: feat: Implement deserialization between `ProgramJson` and `Program`
## Description

The `Program` structure that we currently have in the code actually corresponds to a program coming directly from a JSON so it is rather the `ProgramJson` equivalence that exists in Rust:

https://github.com/lambdaclass/cairo-vm/blob/8a2ef24c1ca258f1d1d230cbbd1478cd1d4cf9db/vm/src/serde/deserialize_program.rs#L74-L85

But to have a smooth and efficient implementation running in the Cairo runner, we need to:
- Rename the existing structure to `ProgramJson`.
- Create another `Program` structure which will be called in the Cairo Runner.
- Create a deserialization method to go from `ProgramJson` to `Program`.

tcoratger:
Should close #256.


cairo_programs/fibonacci.json
```diff
@@ -1,717 +1,684 @@
...
+      }
+    }
+  },
+  "hints": {},
+  "identifiers": {
+    "__main__.fib": {
+      "decorators": [],
+      "pc": 11,
+      "type": "function",
+      "value": 11111111
```
tcoratger, line 461: @jobez @lana-shanghai Here I had to put a random value as I had to do in other places because the parsing notes an error if there is no `value`. Here is the reference for this error during parsing,

https://github.com/lambdaclass/cairo-vm/blob/e6171d66a64146acc16d5512766ae91ae044f297/vm/src/serde/deserialize_program.rs#L479


Is there something I misunderstood, should we put a zero value here when there is nothing?

jobez: kk, per discussion, this error is on the case where the identifier is a const, so we can undo these changes

src/hint_processor/hint_processor_def.zig
```diff
@@ -0,0 +1,87 @@
...
+            .ap_tracking_data = null,
+            .cairo_type = null,
+        };
+    }
+
+    /// Initializes a simple hint reference with the specified offset.
+    ///
+    /// Params:
+    ///   - `offset1`: First offset value for the hint reference.
+    pub fn init_simple(offset1: i32) Self {
```
jobez, line 52: camel casing pls

src/vm/types/programjson.zig
```diff
@@ -0,0 +1,654 @@
...
+        /// File contents associated with debug information.
+        file_contents: json.ArrayHashMap([]const u8),
+        /// Instruction locations linked with debug information.
+        instruction_locations: json.ArrayHashMap(InstructionLocation),
+    },
+    /// Hints associated with the program.
+    hints: json.ArrayHashMap([]const HintParams),
+    /// Identifiers within the program.
+    identifiers: json.ArrayHashMap(Identifier),
+    // identifiers: std.StringHashMap(Identifier),
```
jobez, line 279: can be removed safely :)

src/vm/types/programjson.zig
```diff
@@ -0,0 +1,654 @@
...
+        }
+
+        var instruction_locations = std.StringHashMap(InstructionLocation).init(allocator);
+        errdefer instruction_locations.deinit();
+        // Populates the instruction_locations hashmap with debug information related to instruction locations.
+        for (self.debug_info.instruction_locations.map.keys(), self.debug_info.instruction_locations.map.values()) |key, value| {
+            try instruction_locations.put(key, value);
+        }
+
+        var identifiers = std.StringHashMap(Identifier).init(allocator);
```
jobez, line 410: the underlying impl of the json hashmap is `StringArrayHashMapUnmanaged`: StringArrayHashMapUnmanaged

so I am curious if we can clone here
https://github.com/ziglang/zig/blob/master/lib/std/array_hash_map.zig#L385

jobez (commented):
some comments!

src/hint_processor/hint_processor_def.zig
tcoratger: done

src/vm/types/programjson.zig
tcoratger: done

src/vm/types/programjson.zig
tcoratger: So yes I looked a little in fact we could do that but:
- We would have to go from a simple hashmap to an arrayHashMap and I don't think we necessarily need it.
- We would have to have an unmanaged arrayHashMap, which is a little more complicated to handle afterwards.

So I told myself that the simplest solution was to do it like this but I agree that it's a bit repetitive to go through all the keys and values again.

What do you think?

jobez (approved):
🎉

PR #266, opened 2023-12-16
refactor relocateTrace unit tests

tcoratger:
Now that `TraceContext.Entry` fields are values and not pointers, it is possible to refactor our current complex `relocateTrace` unit test declarations for something less verbose.


lana-shanghai (approved):
lgtm

PR #268, opened 2023-12-19
add builtin init to runner

lana-shanghai:


src/vm/runners/cairo_runner.zig
```diff
@@ -174,6 +196,7 @@ pub const CairoRunner = struct {
         // self.program.deinit();
         self.function_call_stack.deinit();
         self.instructions.deinit();
+        self.vm.builtin_runners.deinit();
```
jobez, line 199: CairoVM handles the builtin_runner management, so this would result in a double free

see:
https://github.com/keep-starknet-strange/cairo-zig/blob/main/src/vm/core.zig#L118

src/vm/types/program.zig
```diff
@@ -175,6 +175,16 @@ pub const Program = struct {
         }
         return parsed_data;
     }
+
+    pub fn readBuiltins(self: Self, allocator: Allocator) !std.ArrayList([]const u8) {
```
jobez, line 179: you can access the `builtins` field directly off parsed_json.value.builtins and it will be a list of strings

src/vm/types/layout.zig
```diff
@@ -153,6 +166,90 @@ pub const CairoLayout = struct {
...
+        for (program_builtins) |builtin| {
+            const case = std.meta.stringToEnum(Case, builtin) orelse return RunnerError.BuiltinNotInLayout;
+            switch (case) {
+                .output => if (self.builtins.output) {
+                    try builtin_runners.append(BuiltinRunner{ .Output = OutputBuiltinRunner.initDefault(allocator) });
+                },
+
+                // TODO: implement initDefault for the rest of the builtin runners.
+
+                .pedersen => if (self.builtins.pedersen != null) {
```
jobez, line 213: there is a school of thought that says writing code in a 'return early' style with the nonhappy path is easier to read. i tend to agree. see: https://medium.com/swlh/return-early-pattern-3d18a41bba8

if we were to apply this principle here we could do something like:

```
fn isBuiltinAvailable(self: *Self, case: Builtin) bool {
    return switch (case) {
        .output => self.builtins.output != null,
        .pedersen => self.builtins.pedersen != null,
        .range_check => self.builtins.range_check != null,
        .ecdsa => self.builtins.ecdsa != null,
        .bitwise => self.builtins.bitwise != null,
        .ec_op => self.builtins.ec_op != null,
        .keccak => self.builtins.keccak != null,
        .poseidon => self.builtins.poseidon != null,
    };
}
```

then we can have early check if we are in the failing case with a return

`if (!self.isBuiltinAvailable(case)) return RunnerError.BuiltinNotInLayout;`

and another switch that handles constructing/appending the right builtin

lmk what you think

jobez (commented):
smol comments

src/vm/runners/cairo_runner.zig
```diff
@@ -28,20 +41,34 @@ pub const CairoRunner = struct {
...
+            .small => runner_layout = CairoLayout.smallInstance(),
+            .dynamic => runner_layout = CairoLayout.dynamicInstance(),
+            .all_cairo => runner_layout = CairoLayout.allCairoInstance(allocator) catch |err| {
+                return err;
+            },
+        }
         return .{
             .allocator = allocator,
             .program = program,
+            .layout = runner_layout,
```
tcoratger, line 71: Here I think you can directly have:

```zig
.layout =  switch (std.meta.stringToEnum(Case, layout) orelse return CairoRunnerError.InvalidLayout) {
            .plain => CairoLayout.plainInstance(),
            .small => CairoLayout.smallInstance(),
            .dynamic => CairoLayout.dynamicInstance(),
            .all_cairo => try CairoLayout.allCairoInstance(allocator)
        }
```

src/vm/types/layout.zig
```diff
@@ -153,6 +166,90 @@ pub const CairoLayout = struct {
...
+        proof_mode: bool,
+        program_builtins: []const []const u8,
+    ) !ArrayList(BuiltinRunner) {
+        var builtin_runners = ArrayList(BuiltinRunner).init(allocator);
+
+        if (proof_mode) {
+            // TODO: initialize all the builtins in the layout.
+        }
+
+        const Case = enum {
```
tcoratger, line 193: We already have a `BuiltinName` enum in `program.zig` file :)

src/vm/runners/cairo_runner.zig
```diff
@@ -50,11 +77,17 @@ pub const CairoRunner = struct {
...
     pub fn initBuiltins(self: *Self, vm: *CairoVM) !void {
-        _ = self;
-        _ = vm;
+        const builtin_runners = try CairoLayout.setUpBuiltinRunners(
+            self.layout,
+            self.allocator,
+            self.proof_mode,
+            self.program.builtins,
+        );
+        vm.builtin_runners = builtin_runners;
```
tcoratger, line 86: Maybe we can directly do

```zig
vm.builtin_runners = try CairoLayout.setUpBuiltinRunners(
            self.layout,
            self.allocator,
            self.proof_mode,
            self.program.builtins,
        );
```

no?

src/vm/types/layout.zig
lana-shanghai: oops :D

src/vm/runners/cairo_runner.zig
lana-shanghai: hashtag idiomatic Zig :)

src/vm/types/layout.zig
lana-shanghai: Great suggestion, improved readability a ton

jobez (approved):
Lgtm

Issue #264, opened 2023-12-15
feat: initialize segments for the builtins inside the cairo runner 


lana-shanghai:
Add the segments initialization in the [Cairo Runner](https://github.com/keep-starknet-strange/ziggy-starkdust/blob/9cd2739b9a4dc478ec41835c606cfe4e267d94cd/src/vm/runners/cairo_runner.zig#L73):

```
    /// Initializes common segments for the execution of a cairo program.
    pub fn initSegments(self: *Self) !void {

        // Common segments, as defined in pg 41 of the cairo paper
        // stores the bytecode of the executed Cairo Program
        self.program_base = try self.vm.segments.addSegment();
        // stores the execution stack
        self.execution_base = try self.vm.segments.addSegment();

        // TODO, add builtin segments when fib milestone is completed
    }
``` 

Follow the Rust [implementation](https://github.com/lambdaclass/cairo-vm/blob/8a2ef24c1ca258f1d1d230cbbd1478cd1d4cf9db/vm/src/vm/runners/cairo_runner.rs#L364) here.

PR #272, opened 2023-12-20
feat/builtin segments and stack initialized in runner 

jobez:
closes: #264 

slightly coupled with #268, but PRing separately because as I was comparing output with the cairovm in rust I noticed that our initial fp wasn't lined up, and it was coupled with having a builtin's stack append the runners, which shifts the initial fp. 

can add test coverage  to the `builtin_runner` in subsequent pr


lana-shanghai (approved):
lgtm

PR #249, opened 2023-12-09
feat: enchance pedersen instance def

dhruvkelawala:
closes #217

PR #278, opened 2023-12-21
Li/feat/init_default range check builtin

lana-shanghai:
In order to run other types of builtin we have to have a `initDefault` function implemented for the different types of runners


src/vm/builtins/builtin_runner/range_check.zig
```diff
@@ -78,6 +78,10 @@ pub const RangeCheckBuiltinRunner = struct {
         };
     }
 
+    pub fn initDefault() Self {
```
jobez, line 81: it would be good to have a bit of a huddle/discussion about how we want to initialize builtins w/ the instance type defs 

my proposal/take:

smol refactor to `range_check_instance_def` where
-  we use default field values: https://ziglang.org/documentation/master/#Default-Field-Values
-  remove current init
-  `from` refactored to `init`
- initialize RangeCheckInstanceDef struct and use ratio/n_parts fields in defaultInit

lmk if this is clear/sounds good 🦆

jobez (commented):
huddle req around a comment, just in time for the holidays

src/vm/builtins/builtin_runner/range_check.zig
dhruvkelawala: Sounds good. But what do you mean by ?

> from refactored to init

dhruvkelawala: Also, in terms of general practice, I also like this point 
https://github.com/keep-starknet-strange/ziggy-starkdust/pull/249#discussion_r1429181712

jobez: > Sounds good. But what do you mean by ?
> 
> > from refactored to init

Great! Here I meant specifically for 

https://github.com/keep-starknet-strange/ziggy-starkdust/blob/c65c9dd984e05aef33909bc99bd9b0628d5043eb/src/vm/types/range_check_instance_def.zig#L34-L39

we would refactor `from` to `init` here

dhruvkelawala: Ohh yes, that's great! Sounds good to me

jobez (approved):

PR #283, opened 2023-12-22
Added relocate and relocate memory

### LINKED ISSUE #231: feat: implement relocate and relocateMemory
The relocation logic is made up of a couple of functions:

- `relocationSegments`
- `relocateTrace`
- `relocateMemory`
- `relocate` 

The first two have been implemented here #156

Godspower-Eze:
Closes #231


src/vm/memory/segments.zig
```diff
@@ -250,6 +250,44 @@ pub const MemorySegmentManager = struct {
...
+        ).init(allocator);
+        errdefer relocated_memory.deinit();
+        for (0..self.memory.num_segments) |i| {
+            const segment_index = @as(i64, @intCast(i));
+            const size = self.getSegmentSize(@intCast(i));
+            if (size != null) {
+                for (0..size.?) |j| {
+                    const key = Relocatable.init(segment_index, @as(u32, @truncate(j)));
+                    const cell = self.memory.get(key);
+                    if (cell != null) {
```
tcoratger, line 280: Here you can use:

```zig
if (cell) |c| {
   ...
}
```

src/vm/memory/segments.zig
```diff
@@ -950,6 +988,68 @@ test "MemorySegmentManager: relocateSegments for ten segments" {
...
+    const fourth = expected_relocated_memory.get(4);
+    const fourth_actual = relocated_memory.get(4);
+
+    const fifth = expected_relocated_memory.get(5);
+    const fifth_actual = relocated_memory.get(5);
+
+    const sixth = expected_relocated_memory.get(9);
+    const sixth_actual = relocated_memory.get(9);
+
+    try expectEqual(first, first_actual);
```
tcoratger, line 1045: Here in my opinion, you don't need to create an `expected_relocated_memory` `AutoHashMap`, you can directly check:
```zig
try expectEqual(Felt252.fromInteger(656720927189829), relocated_memory.get(1));
```

I think that you should also test the total number of elements in `relocated_memory` by using the `count()` method in order to confirm that you have the expected number of elements there.

Godspower-Eze: I think laying it out that is good for better understanding the test and it makes it look less like a magic value. What do you think?

jobez (approved):
per discussion, approve merging this with the follow up refactor issue as defined in #305

PR #293, opened 2023-12-24
feat: add `outerBrackets` method for parsing and unit test

tcoratger:
## Description

This PR introduces the `outerBrackets` method to parse outermost brackets within a byte array. Additionally, it includes a corresponding unit test to ensure the functionality works as expected.

## Changes Made

- Added `outerBrackets` method for parsing outer brackets in a byte array.
- Included a unit test suite to validate the functionality of the new method.


src/parser/deserialize_utils.zig
```diff
@@ -0,0 +1,82 @@
...
+/// If no brackets are found or if the brackets are improperly formatted, it returns a `ParseOptResult`
+/// containing the original `input` and a `false` boolean value.
+///
+/// # Parameters
+/// - `input`: A byte array in which outer brackets need to be parsed.
+///
+/// # Returns
+/// A `ParseOptResult` tuple containing the content within the outer brackets (if found) and a boolean
+/// indicating successful parsing.
+pub fn outerBrackets(input: []const u8) ParseOptResult {
```
jobez, line 26: what do you think about splitting up the cases? i find it makes it easier to read. comments are just for explanation of the smol restructuring, feel free to keep your own.

```
pub fn outerBrackets(input: []const u8) ParseOptResult {
    var it_in = std.mem.splitSequence(u8, input, "[");
    var it_out = std.mem.splitBackwardsSequence(u8, input, "]");

    // explicit check if indexes are null
    if (it_in.index == null and it_out.index == null) {
        // no brackets
        return .{ input, false };
    }

    // then we refine to that search was a match and it is the beginning and end of the string
    if (std.mem.eql(u8, it_in.first(), "") and std.mem.eql(u8, it_out.first(), "")) {
        return .{ input[it_in.index.?..it_out.index.?], true };
    }

    // If the above conditions are not met, return original input with false
    return .{ input, false };
}

```

jobez (commented):
other than small nit about structure of the logic, lgtm.

src/parser/deserialize_utils.zig
tcoratger: Thanks yeah evaluated this possibility but it sucks for the following case

```zig
 // Test case where input is an empty string
    const empty_value = outerBrackets("");
```

In this scenario if I’ve

```zig
pub fn outerBrackets(input: []const u8) ParseOptResult {
    // Split the input array at each '[' character
    var it_in = std.mem.splitSequence(u8, input, "[");

    // Split the input array at each ']' character, searching backward
    var it_out = std.mem.splitBackwardsSequence(u8, input, "]");

    std.debug.print("Iterator in before mem check = {any}\n", .{it_in});
    std.debug.print("Iterator out before mem check = {any}\n", .{it_out});

    // Check if the first element after splitting at '[' and ']' is an empty string
    if (std.mem.eql(u8, it_in.first(), "") and std.mem.eql(u8, it_out.first(), "")) {
        std.debug.print("Iterator in after mem check = {any}\n", .{it_in});
        std.debug.print("Iterator out after mem check = {any}\n", .{it_out});

        // Check if both 'it_in' and 'it_out' indices are not null
        if (it_in.index != null and it_out.index != null) {
            // Return a tuple containing the content within the outer brackets and true
            return .{ input[it_in.index.?..it_out.index.?], true };
        }
    }

    // Return the original input and a false boolean value indicating unsuccessful parsing
    return .{ input, false };
}
```

I’ve as a result
```shell
Iterator in before mem check = mem.SplitIterator(u8,.sequence){ .buffer = {  }, .index = 0, .delimiter = { 91 } }
Iterator out before mem check = mem.SplitBackwardsIterator(u8,.sequence){ .buffer = {  }, .index = 0, .delimiter = { 93 } }
Iterator in after mem check = mem.SplitIterator(u8,.sequence){ .buffer = {  }, .index = null, .delimiter = { 91 } }
Iterator out after mem check = mem.SplitBackwardsIterator(u8,.sequence){ .buffer = {  }, .index = null, .delimiter = { 93 } }
```
so that it's impossible to catch the null index before moving the iterator first.

In fact the null check was to handle the empty string stuff. So I just rearranged all of that and check at the beginning of the function the special case of empty string. Let me know what you think.

lana-shanghai (approved):
lgtm

PR #310, opened 2024-01-11
port:native pedersen hash func

### LINKED ISSUE #309: porting: pedersen hashing func on native zig
porting pedersen hashing func based on rust implementation
[pederesen hash as implemented in starknet-rs](https://github.com/xJonathanLEI/starknet-rs/blob/master/starknet-crypto/src/pedersen_hash.rs)

StringNick:
closes #309


src/math/crypto/curve/ec_point.zig
```diff
@@ -0,0 +1,150 @@
...
+
+pub const ProjectivePoint = struct {
+    const Self = @This();
+
+    x: Felt252,
+    y: Felt252,
+    z: Felt252,
+    infinity: bool,
+
+    pub fn from_affine_point(p: AffinePoint) Self {
```
tcoratger, line 15: The convention for zig functions at the moment is to put camel case in the function names. It's sometimes strange...

image

src/math/crypto/curve/ec_point.zig
```diff
@@ -0,0 +1,150 @@
...
+
+        const uy = u.mul(self.y);
+
+        const x = u.mul(w);
+        const y = t.mul(v.sub(w)).sub(Felt252.two().mul(uy).mul(uy));
+        const z = u.mul(u).mul(u);
+
+        self.x = x;
+        self.y = y;
+        self.z = z;
```
tcoratger, line 43: The following should work no?

```zig
self.* = .{
.x = u.mul(w),
.y = t.mul(v.sub(w)).sub(Felt252.two().mul(uy).mul(uy)),
.z = u.mul(u).mul(u),
}

Comment (StringNick) [reply src/math/crypto/curve/ec_point.zig:43]: should be

File: src/math/crypto/curve/ec_point.zig [hunk 3]
```diff
@@ -0,0 +1,149 @@
...
+    y: Felt252,
+    z: Felt252,
+    infinity: bool,
+
+    pub fn fromAffinePoint(p: AffinePoint) Self {
+        return .{
+            .x = p.x,
+            .y = p.y,
+            .z = Felt252.one(),
+            .infinity = false,
```
tcoratger, line 20: If you apply the previous default suggestions, `.z` and `.infinity` are not needed anymore.

src/math/crypto/curve/ec_point.zig
```diff
@@ -0,0 +1,149 @@
...
+        // l = (3x^2+a)/2y with a=1 from stark curve
+        const lambda = Felt252.three().mul(self.x.mul(self.x)).add(Felt252.one()).mul(Felt252.two().mul(self.y).inv().?);
+
+        const result_x = lambda.mul(lambda).sub(self.x).sub(self.x);
+        self.y = lambda.mul(self.x.sub(result_x)).sub(self.y);
+        self.x = result_x;
+    }
+
+    pub fn fromProjectivePoint(p: *ProjectivePoint) Self {
+        const zinv = p.z.inv().?;
```
tcoratger, line 141: Here you will have an error you need to handle in the case `p.z.inv()` is null because you will try to access a null field, returning a panic.

src/math/crypto/pedersen/pedersen.zig
```diff
@@ -0,0 +1,79 @@
...
+
+    return result;
+}
+
+fn addPoints(acc: *ProjectivePoint, bits: []const bool, prep: []const AffinePoint) void {
+    // Preprocessed material is lookup-tables for each chunk of bits
+    const table_size = (1 << CURVE_CONSTS_BITS) - 1;
+
+    var i: usize = 0;
+    while (i < bits.len / CURVE_CONSTS_BITS) : (i += 1) {
```
tcoratger, line 33: why not using

```zig
for (0..bits.len / CURVE_CONSTS_BITS) |i|  {
...
}
```

in order to avoid declaring a `i` variable

src/math/fields/fields.zig
```diff
@@ -209,6 +209,22 @@ pub fn Field(
             return ret;
         }
 
+        /// Convert the field element to a bits little endian array.
+        ///
+        /// This function converts the field element to a byte array for serialization.
+        pub fn toBitsLe(self: Self) [@bitSizeOf(u256)]bool {
```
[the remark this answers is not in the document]
StringNick: this one converts to bit for hashing func where we use convert from bits as in rust implementation

src/math/crypto/curve/ec_point.zig
StringNick: as i understand it should not be null (as rust implementation they unwrap also, expecting not null), should we handle null value?

src/math/fields/fields.zig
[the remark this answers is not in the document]
tcoratger: Ok fine, then fine :)

src/math/crypto/curve/ec_point.zig
tcoratger: I think so just in case, maybe throwing an error or something like that to avoid panic, wdt?

StringNick: maybe use some like return unreachble?

lana-shanghai: maybe smth like `if (p.z.inv) |zinv| { doStuff } else { ... } ` , maybe `unreachable` is actually best if you're sure there is never a null value there?

PR #311, opened 2024-01-11
EC OP builtin runner

lana-shanghai:


src/vm/builtins/builtin_runner/ec_op.zig
```diff
@@ -10,6 +11,14 @@ const MemorySegmentManager = @import("../../memory/segments.zig").MemorySegmentM
 const AutoHashMap = std.AutoHashMap;
 const ArrayList = std.ArrayList;
 const Allocator = std.mem.Allocator;
+const Tuple = std.meta.Tuple;
+
+const EC_POINTS = [_]Tuple(&.{ usize, usize }){
```
tcoratger, line 16: Why not using an array of `EcPoint` struct directly here? In order to represent the coordinates in a proper way and to avoid the usage of tuples.

src/vm/builtins/builtin_runner/ec_op.zig
```diff
@@ -66,26 +75,65 @@ pub const EcOpBuiltinRunner = struct {
...
+            if (memory.get(try instance.addFelt(Felt252.fromInteger(i)))) |cell| {
+                const felt = try cell.tryIntoFelt();
+                try input_cells.append(felt);
+            }
+        }
+
+        for (EC_POINTS[0..2]) |pair| {
+            const x = input_cells.items[pair[0]];
+            const y = input_cells.items[pair[1]];
+            var point = EC.ECPoint{ .x = x, .y = y };
```
tcoratger, line 114: I think we should add an `init` method to `ECPoint` struct in order to avoid this kind of raw initialization wdt?

lana-shanghai: Yes, I will refactor this to use the AffinePoint struct introduced by Nikita

tcoratger (approved):

PR #312, opened 2024-01-12
Normalize conventions around instance type definitions across codebase

### LINKED ISSUE #302: Normalize conventions around instance type definitions across codebase
Per huddle/consensus established [here](https://github.com/keep-starknet-strange/ziggy-starkdust/pull/278#discussion_r1435157767_), there is atleast one instance def (range_check_instance_def) that should be refactored by:

-  using default field values: https://ziglang.org/documentation/master/#Default-Field-Values rather than a defaultInit method, unless there are specific fields that need allocation
-  remove current init
-  `from` refactored to `init`
- initialize RangeCheckInstanceDef struct and use ratio/n_parts fields in defaultInit

tcoratger:
Should close #302.

### DIFF
--- a/src/vm/types/bitwise_instance_def.zig
+++ b/src/vm/types/bitwise_instance_def.zig
@@ -1,8 +1,11 @@
+const std = @import("std");
+const expectEqual = std.testing.expectEqual;
+
 /// Number of cells by bitwise operation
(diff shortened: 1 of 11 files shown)


jobez (approved):
excellent!

PR #321, opened 2024-01-19
Feat/decode current instructions

### LINKED ISSUE #282: feat: implement `decodeCurrentInstruction` method for `CairoVM`
## Description

In order to be able to decode our current instruction defined by the VM run context program counter, we need to implement a `decodeCurrentInstruction` method for `CairoVM`.

## References
- Reference for the Rust implementation: https://github.com/lambdaclass/cairo-vm/blob/8a2ef24c1ca258f1d1d230cbbd1478cd1d4cf9db/vm/src/vm/vm_core.rs#L434-L442

dhruvkelawala:
closes #282


src/vm/decoding/decoder.zig
```diff
@@ -0,0 +1,275 @@
...
+const ApUpdate = Instructions.ApUpdate;
+
+const expectError = std.testing.expectError;
+const expectEqual = std.testing.expectEqual;
+
+//  0|  opcode|ap_update|pc_update|res_logic|op1_src|op0_reg|dst_reg
+// 15|14 13 12|    11 10|  9  8  7|     6  5|4  3  2|      1|      0
+
+/// Decodes an instruction. The encoding is little endian, so flags go from bit 63 to 48.
+pub fn decodeInstructions(encoded_instr: u64) !Instruction {
```
tcoratger, line 20: In think we already have this function out there no?

https://github.com/keep-starknet-strange/ziggy-starkdust/blob/842ff1acf8598dd89901ea78f6453e899504f8fd/src/vm/instructions.zig#L197-L263

I really like your constant declarations because they are much more explicit than what is currently done with harcoded values, perhaps you could replace the existing function with yours which is clearer?

Also look if there are unit tests that you have done that are not done in the current implementation, what do you think?

dhruvkelawala: Isn't it a good idea to just remove/cleanup the old function and have this dedicated file for decoder?

tcoratger: Yes for me we can do that, what do you think @lana-shanghai ?

lana-shanghai: Yes, I think a decoder can live in its own file. So you would move all the parsing of opcodes and so on there as well?

lana-shanghai (approved):
lgtm

PR #325, opened 2024-01-19
use default field values for builtins

tcoratger:
Related to this convention: https://ziglang.org/documentation/master/#Default-Field-Values, applied for builtins.


lana-shanghai (approved):
lgtm

Issue #329, opened 2024-01-25
feat: implement ResourceTracker interface and RunResources struct


jobez:
# Description

[ResourceTracker](https://github.com/lambdaclass/cairo-vm/blob/main/vm/src/vm/runners/cairo_runner.rs#L96), as defined in the cairo vm in rust, is a trait `mostly used in the context of starknet, where contracts can call other contracts while sharing the same step limit` (per impl doc string).

[RunResources](https://github.com/lambdaclass/cairo-vm/blob/main/vm/src/vm/runners/cairo_runner.rs#L113-L142) is a struct defining the `n_steps` available to be tracked, as well as its own implementation of ResourceTracker.

In practice, both ResourceTracker and RunResources is used by implementations of HintProcessors, see: 
- within the cairovm in rust
   - [builtin processor being constructed with RunResources::default](https://github.com/lambdaclass/cairo-vm/blob/main/vm/src/hint_processor/builtin_hint_processor/builtin_hint_processor_definition.rs#L161)
   - [builtin hint processor implementing ResourceTracker](https://github.com/lambdaclass/cairo-vm/blob/main/vm/src/hint_processor/builtin_hint_processor/builtin_hint_processor_definition.rs#L834-L850)
   - [cairo 1 hint processor implementing ResourceTracker](https://github.com/lambdaclass/cairo-vm/blob/main/vm/src/hint_processor/cairo_1_hint_processor/hint_processor.rs#L1116-L1132)
- within Blockifier, which depends/includes the cairovm's types and trait as dependency:
    -  https://github.com/starkware-libs/blockifier/blob/54f2f5b445404963e1f91252d96f751d36eba58d/crates/blockifier/src/execution/syscalls/hint_processor.rs#L570-L586

Given these concrete usage patterns, one must consider which of the [options zig has to have interfaces](https://zig.news/yglcode/code-study-interface-idiomspatterns-in-zig-standard-libraries-4lkj) is best suited.

PR #333, opened 2024-01-26
feat: RunResources and ResourceTracker interface

jobez:
This PR resolves issue #329.

**Rationale for Choosing Dynamic Dispatch Approach:**
- **Tagged Enums Limitation:** While tagged enums are effective for a closed set of cases within a single codebase, they are less suitable for scenarios requiring extensibility. HintProcessors often need to be implemented and extended by external systems using the Cairo VM as a library. An example can be seen in the [Blockifier implementation](https://github.com/starkware-libs/blockifier/blob/54f2f5b445404963e1f91252d96f751d36eba58d/crates/blockifier/src/execution/syscalls/hint_processor.rs#L570-L586).
- **`comptime` Constraints:** Utilizing a `comptime` generic interface imposes a limitation: all details about resource monitoring would need to be fully known at compile time. This is impractical for the dynamic nature of HintProcessors, where such details often depend on runtime conditions.
- **Adopting the Fat Pointer Convention:** To address these challenges, we've adopted the 'fat pointer' approach, similar to how allocators are implemented in Zig. This method allows for greater flexibility and runtime polymorphism, enabling external systems to extend or customize HintProcessor implementations without constraints on compile-time knowledge. See [here](https://pithlessly.github.io/allocgate.html) for a writeup of zig's transition of the allocator's approach to polymorphism to fatpointers


src/vm/runners/cairo_runner.zig
```diff
@@ -26,6 +26,76 @@ const expectEqual = std.testing.expectEqual;
 const expectError = std.testing.expectError;
 const expectEqualSlices = std.testing.expectEqualSlices;
 
+/// Tracks the step resources of a cairo execution run.
+const RunResources = struct {
```
tcoratger, line 30: Shouldn't we add a `init` method here to init a `RunResources` instance from a number of steps:

```zig
pub fn consumed(n_steps: usize) Self {
    return .{.n_steps = n_steps};
}
```

wdt?

src/vm/runners/cairo_runner.zig
```diff
@@ -26,6 +26,76 @@ const expectEqual = std.testing.expectEqual;
...
+    // We consider the 'default' mode of RunResources having infinite steps.
+    n_steps: ?usize = null,
+
+    pub fn consumed(self: *RunResources) bool {
```
tcoratger, line 34: Maybe we could define a `Self` type at the beginning of the struct to avoid repetitions of structure name, wdt?

src/vm/runners/cairo_runner.zig
```diff
@@ -26,6 +26,76 @@ const expectEqual = std.testing.expectEqual;
...
+    // define interface fields: ptr,vtab
+    ptr: *anyopaque, //ptr to instance
+    vtab: *const VTab, //ptr to vtab
+    const VTab = struct {
+        consumed: *const fn (ptr: *anyopaque) bool,
+        consumeStep: *const fn (ptr: *anyopaque) void,
+    };
+
+    /// Returns true if there are no resource-steps available.
+    pub fn consumed(self: ResourceTracker) bool {
```
tcoratger, line 63: Maybe we could define a `Self` type at the beginning of the struct to avoid repetitions of structure name, wdt?

tcoratger (approved):

Issue #248, opened 2023-12-08
test: enhance `initSegments` and `initState` implementation for Cairo runner


tcoratger:
### Current Status
Presently, the `initSegments` and `initState` methods within our Cairo runner lack a comprehensive implementation and are devoid of unit tests. This gap in functionality and testing is evident in the following code section:

https://github.com/keep-starknet-strange/ziggy-starkdust/blob/3dc35d39e4f7de477f1c9faa625a8b7e2ce86337/src/vm/runners/cairo_runner.zig#L60-L92

### Objectives
To bolster the reliability and robustness of our codebase, it's imperative to fortify these implementations and augment them with the necessary unit tests. This enhancement will ensure the secure and stable functioning of these critical methods within the Cairo runner.

### Action Plan
1. **Implementation Enhancement:** Address the incomplete sections of the `initSegments` and `initState` methods to achieve a more comprehensive and refined implementation (we need to skip the builtin initialization at the moment and leave it as TODO).
2. **Unit Testing:** Introduce rigorous unit tests to validate the correctness and reliability of these implementations, ensuring they adhere to expected behavior and edge cases.

### References and Guidance
For guidance and inspiration, we will reference the following:

- **Rust File Reference:** [Cairo runner in Rust](https://github.com/lambdaclass/cairo-vm/blob/8887e60f7ea5459188cb6624acc908ad29e75f60/vm/src/vm/runners/cairo_runner.rs) - This serves as a benchmark for unit tests and implementation in Rust.
- **Python File Reference:** [Cairo runner unit tests in Python](https://github.com/starkware-libs/cairo-lang/blob/ee7ce74e1159a349d4b77a5f952241b50b1692de/src/starkware/cairo/lang/vm/cairo_runner_test.py) - Offering insights into the structure and testing paradigm in Python for comparison and reference.

tcoratger:
Will work on this

PR #338, opened 2024-01-29
improve init methods in `CairoRunner` with unit tests

tcoratger:
## Description
This pull request enhances the initialization methods (`init`) in the `CairoRunner` module by refining the logic and adding comprehensive unit tests for better coverage.

Related #248.

## Changes Made
- Updated the `init` methods in the `CairoRunner` module to improve clarity, maintainability, and consistency.
- Added a set of unit tests to thoroughly cover the initialization process and ensure correct behavior.

### DIFF
--- a/src/vm/runners/cairo_runner.zig
+++ b/src/vm/runners/cairo_runner.zig
@@ -167,7 +167,7 @@ pub const CairoRunner = struct {
         try self.initBuiltins(&self.vm);
         try self.initSegments(null);
         const end = try self.initMainEntrypoint();
-        self.initVM();
+        try self.initVM();
         return end;
     }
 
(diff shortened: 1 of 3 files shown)


src/vm/runners/cairo_runner.zig
```diff
@@ -192,7 +204,7 @@ pub const CairoRunner = struct {
     /// - `entrypoint:` The address, relative to the program segment, where execution begins.
     pub fn initState(self: *Self, entrypoint: usize) !void {
         self.initial_pc = self.program_base;
-        self.initial_pc.addUintInPlace(entrypoint);
+        self.initial_pc.?.addUintInPlace(entrypoint);
```
lana-shanghai, line 207: Is it safe to unwrap, do we need a try here maybe?

tcoratger: No it is not safe indeed but this function is not completed and tested (previously we had potential `undefined` here). So I just replaced by an optional type and as a default I put `.?.` because we need to handle this error properly. Will open a PR to complete and test this function

tcoratger: It is also planned to modify that as a second part of https://github.com/keep-starknet-strange/ziggy-starkdust/issues/248.

lana-shanghai (approved):
lgtm

PR #340, opened 2024-01-30
add `parseOffset` and `parseRegister` to deserialization utils

tcoratger:
### Description
This pull request introduces two new utility functions, `parseOffset` and `parseRegister`, to enhance deserialization capabilities. These functions facilitate the parsing of offsets and registers from byte arrays, providing additional tools for handling deserialization tasks.

### Changes
- Added `parseOffset` function to parse numeric offsets from byte arrays.
- Added `parseRegister` function to identify and parse registers ("fp" and "ap") from byte arrays.

### Usage
These utility functions can be leveraged to extract offset values and register information during deserialization processes.

### Warning
These parsing functions assume that the input text is well formed (`ProgramJson`). Malformed Json can lead to unexpected parsing errors. In the future, we may want to raise specific errors in these cases if we see that it is necessary.

### DIFF
--- a/src/parser/deserialize_utils.zig
+++ b/src/parser/deserialize_utils.zig
@@ -1,4 +1,5 @@
 const std = @import("std");
+const Register = @import("../vm/instructions.zig").Register;
 
 const expect = std.testing.expect;
 const expectEqual = std.testing.expectEqual;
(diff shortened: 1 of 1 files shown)


lana-shanghai (approved):
lgtm

PR #341, opened 2024-01-31
implement HintProcessorData

jobez:
## Description
This pull request adds the `HintProcessorData` type along with the `getIdsData` helper function that is used in the `compileHint` method of the HintProcessor interface

## Notable differences from porting from the rust impl

- We implement the type in the `hint_processor_def` as opposed to defining it within the `builtin_hint_processors_def` to avoid circular dependences.

### DIFF
--- a/src/hint_processor/hint_processor_def.zig
+++ b/src/hint_processor/hint_processor_def.zig
@@ -1,4 +1,9 @@
 const std = @import("std");
+const Allocator = std.mem.Allocator;
+const ArrayList = std.ArrayList;
+const StringHashMap = std.StringHashMap;
+
+const CairoVMError = @import("../vm/error.zig").CairoVMError;
 const OffsetValue = @import("../vm/types/programjson.zig").OffsetValue;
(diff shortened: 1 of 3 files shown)


src/hint_processor/hint_processor_def.zig
```diff
@@ -53,6 +94,42 @@ pub const HintReference = struct {
...
+    var ref_id_it = reference_ids.iterator();
+    while (ref_id_it.next()) |ref_id_entry| {
+        const path  = ref_id_entry.key_ptr.*;
+        const ref_id = ref_id_entry.value_ptr.*;
+
+        var name_iterator = std.mem.splitBackwards(u8, path, @as([]const u8, "."));
+        
+        const name = name_iterator.next() orelse return CairoVMError.Unexpected;
+
+        if (ref_id >= references.len) return CairoVMError.Unexpected;
```
tcoratger, line 122: I think you should put this condition at the beginning of the loop to avoid doing the rest of the calculations if this one is not satisfied what do you think?

tcoratger (approved):

PR #380, opened 2024-02-15
Refactor/affinepoint

### LINKED ISSUE #352: refactor: use AffinePoint instead of ECPoint
##Description 

Currently the EC OP builtin [uses](https://github.com/keep-starknet-strange/ziggy-starkdust/blob/main/src/vm/builtins/builtin_runner/ec_op.zig#L25) the [ECPoint](https://github.com/keep-starknet-strange/ziggy-starkdust/blob/main/src/math/fields/elliptic_curve.zig#L22) struct, replace it with the [AffinePoint](https://github.com/keep-starknet-strange/ziggy-starkdust/blob/main/src/math/crypto/curve/ec_point.zig#L93). 

Make sure that the necessary operations such as `[ecDoubleSlope](https://github.com/keep-starknet-strange/ziggy-starkdust/blob/main/src/math/fields/elliptic_curve.zig#L106)` and [`ecOpImpl`](https://github.com/keep-starknet-strange/ziggy-starkdust/blob/main/src/math/fields/elliptic_curve.zig#L153) are implemented for the AffinePoint struct. 

Remove the unnecessary `elliptic_curve` file.

oxlime:
refactor should close #352


src/math/crypto/curve/ec_point.zig
```diff
@@ -236,4 +259,158 @@ pub const AffinePoint = struct {
...
+    var slope = m.toInteger();
+    var partial_sum = const_partial_sum;
+    var doubled_point = const_doubled_point;
+
+    for (0..height) |_| {
+        if (doubled_point.x.sub(partial_sum.x).equal(Felt252.zero())) {
+            return ECError.XCoordinatesAreEqual;
+        }
+        if (slope & 1 != 0) {
+            //     partial_sum = try partial_sum.add(doubled_point);
```
tcoratger, line 349: Can be removed

src/math/crypto/curve/ec_point.zig
```diff
@@ -152,6 +163,18 @@ pub const AffinePoint = struct {
...
 
+    /// Initializes an `AffinePoint` with the specified x and y coordinates.
+    ///
+    /// # Parameters
+    /// - `x`: The x-coordinate of the point.
+    /// - `y`: The y-coordinate of the point.
+    ///
+    /// # Returns
+    /// An initialized `ECPoint` with the provided coordinates.
+    pub fn init(x: Felt252, y: Felt252, infinity: bool) Self {
```
tcoratger, line 174: Here better to call it `initUnchecked` as long as we don't check that the provided coordinates give a point that is on the curve.

src/math/crypto/curve/ec_point.zig
```diff
@@ -236,4 +259,158 @@ pub const AffinePoint = struct {
...
+
+    /// Doubles a point on an elliptic curve with the equation y^2 = x^3 + alpha*x + beta.
+    ///
+    /// # Arguments
+    /// - `self` - The point.
+    /// - `alpha` - The alpha parameter of the elliptic curve.
+    ///
+    /// # Returns
+    /// The doubled elliptic curve point.
+    pub fn ecDouble(self: *Self, alpha: Felt252) ECError!AffinePoint {
```
tcoratger, line 271: Here I would say that it is better to mark `alpha` as comptime as long as we know it as a constant of each curve. Or maybe we can put it as an additional field of the struct.

oxlime: i decided to make it part of the struct because there were errors with having it be comptime for example:
        doubled_point = try doubled_point.ecDouble(alpha);

my issue now is with adding all of the .alpha to the gen/constants.zig

tcoratger (approved):

PR #409, opened 2024-02-27
Add unit test for `getConstants` in Cairo runner

### LINKED ISSUE #400: Cairo runner: add unit tests to get program constants
Now that #392 is merged, we need some unit tests to confirm that program constants can be accessed properly inside the Cairo runner. We don't need any additional function to get that because Cairo runner fields can be publicly accessed but we need some unit tests to confirm the correct access.

Related:
- https://github.com/lambdaclass/cairo-vm/blob/4ea224a1068b43aaa3899d24e834475106e8a78f/vm/src/vm/runners/cairo_runner.rs#L3542-L3550

tcoratger:
Should close #400.


lana-shanghai (approved):
lgtm

PR #434, opened 2024-02-29
Add segment related hint

### LINKED ISSUE #422: BuiltinHint: implement `segments` utils
We need to implement `segments` utils in the context of our builtin hint processor.

Rust reference: https://github.com/lambdaclass/cairo-vm/blob/main/vm/src/hint_processor/builtin_hint_processor/segments.rs

tcoratger:
Should close #422.


lana-shanghai (approved):
lgtm

PR #448, opened 2024-03-09
Implement `endRun` for the Cairo VM

tcoratger:
## Description
This pull request implements the `endRun` function for the Cairo Virtual Machine. The `endRun` function is responsible for finalizing the execution run by verifying auto deductions and marking the run as finished. Additionally, it checks if there is only one execution scope remaining and returns an error if so.

## Changes Made
- Added the `endRun` function to the Cairo VM.
- Implemented the logic to verify auto deductions and mark the run as finished.
- Added checks to handle the case where only one execution scope is remaining.


lana-shanghai (approved):
lgtm

PR #455, opened 2024-03-11
example just for math*.cairo files

jobez:
if i didn't meet the 1:30 target i set for myself, its because it took forever to recompile zig... and the wifi was slow...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,9 @@ $(CAIRO_VM_CLI):
 
 build_cairo_vm_cli: | $(CAIRO_VM_CLI)
 BENCH_DIR=cairo_programs/benchmarks
+# Define the directory containing your Cairo programs
+CAIRO_PROGRAMS_DIR := cairo_programs
+
 
 # Creates a pyenv and installs cairo-lang
 deps:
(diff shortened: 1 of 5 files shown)


cairo_programs/math_cmp.cairo
```diff
@@ -0,0 +1,62 @@
+%builtins range_check
```
jobez, line 1: we commit cairo files. i arbitrarily chose to just move the files in the great https://github.com/lambdaclass/cairo-vm/tree/main/cairo_programs

that matched "math.*cairo*

Makefile
```diff
@@ -39,6 +42,19 @@ test-filter:
 build-integration-test:
 	@zig build integration_test
 
+
+# Task for running the custom integration process
+run-custom-integration-test:
```
jobez, line 47: just for the sake of demonstration i created a separate entry point to run the integration tests that 

- specifically look up all the cairo files matching an arbitary glob
- write all matches to a text file
- compile all matches
- pass the text file into the integration_test runner to actually run

src/integration_tests.zig
```diff
@@ -5,119 +5,176 @@ const CairoVM = @import("vm/core.zig").CairoVM;
...
+    var test_files_list: ?[]const u8 = null;
+    if (args.len > 1) {
+        // TODO: Parse command-line arguments more rigorously with a CLI utility
+        test_files_list = args[1];
+    }
+
+    var cairo_programs: []CairoTestProgram  = undefined; // This will be defined after reading from the test file or defaulting to all tests.
+
+    // If there are specific tests to run, read them from the file.
+    if (test_files_list) |list| {
```
jobez, line 32: right now this is just a dumb 'if', if there's NO command line arg, just run the programs we define in the source. if there is

- separate by new line
- construct a `CairoTestProgram` where we presume `all_cairo` layout (in the https://github.com/lambdaclass/cairo-vm/blob/main/vm/src/tests/cairo_run_test.rs#L9 file this is the single case where it wasn't using the `all_cairo` layout. need to double check tho as there is loud whispering in the library rn)
- run the programs

jobez (commented):
self-review

PR #463, opened 2024-03-13
Add unit tests for hint utils

tcoratger:
## Add unit tests for hint utils

This pull request adds unit tests for the hint utils module to ensure its functionality is thoroughly tested.

### Changes Made
- Added unit tests covering various functions in the hint utils module.
- Verified the behavior of each function with valid and invalid inputs.
- Ensured appropriate error handling and edge case coverage.

### DIFF
--- a/src/hint_processor/hint_utils.zig
+++ b/src/hint_processor/hint_utils.zig
@@ -16,6 +16,19 @@ const Allocator = std.mem.Allocator;
 const HintError = @import("../vm/error.zig").HintError;
 const hint_processor_utils = @import("hint_processor_utils.zig");
 
+/// Retrieves a constant value based on its variable name.
+///
+/// This function searches for a constant with a variable name suffix in the provided hashmap of constants.
+/// It iterates over the hashmap keys, comparing them with the variable name suffix.
+/// If a match is found, it returns the corresponding constant value.
+///
+/// # Parameters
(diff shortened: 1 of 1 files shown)


src/hint_processor/hint_utils.zig
```diff
@@ -341,3 +367,94 @@ test "getPtrFromVarName: invalid" {
...
+    try ids_data.put("value", HintReference.initSimple(-8));
+
+    // Invokes `getPtrFromVarName` to retrieve the pointer from the "value" variable, expecting an error.
+    try expectEqual(
+        HintError.UnknownIdentifier,
+        getRelocatableFromVarName("value", &vm, ids_data, .{}),
+    );
+}
+
+test "getConstantFromVarName: valid without point in the middle of var name" {
```
lana-shanghai, line 423: isn't it the other way around? with / without the period?

lana-shanghai (approved):
lgtm

src/hint_processor/hint_utils.zig
tcoratger: Don't think so because we have to handle the point case which means that for `ids.MAX_LOW` we must capture `MAX_LOW` (which is tested in the following unit test). In this specific test, I am testing the simplest case (without any point in the middle) with only `value` which must be captured correctly. I also added `ids.MAX_LOW` in the hashmap but it is of no use, just to check that it does not pollute the test.

PR #467, opened 2024-03-13
`HintRanges` initialization fix

tcoratger:
In case of non extensive hints, `appendNTimes(null, max_hint_pc + 1)` should not append at init step of the `HintRanges` arrayList in order to avoid spurious preallocation.


lana-shanghai (approved):
lgtm

PR #485, opened 2024-03-21
Math not implemented hints

### LINKED ISSUE #484: BuiltinHint: implement MathUtils
Current math hints taken from golang implementation, so some hints is not implemented, so detect and implement them.
Rust implementation [link](https://github.com/lambdaclass/cairo-vm/blob/main/vm/src/hint_processor/builtin_hint_processor/math_utils.rs) .

StringNick:
closes #484


src/hint_processor/math_utils.zig
```diff
@@ -0,0 +1,685 @@
...
+
+    // assert (2**250 < ADDR_BOUND <= 2**251) and (2 * 2**250 < PRIME) and (
+    //      ADDR_BOUND * 2 > PRIME), \
+    //      'normalize_address() cannot be used with the current constants.'
+    // The second check is not needed, as it's true for the CAIRO_PRIME
+    if (!(lower_bound < addr_bound and addr_bound <= upper_bound or (addr_bound << 1) > STARKNET_PRIME))
+        return HintError.AssertionFailed;
+
+    // Main logic: ids.is_small = 1 if ids.addr < ADDR_BOUND else 0
+    const is_small = if (addr.lt(Felt252.fromInt(u256, addr_bound))) Felt252.one() else Felt252.zero();
```
tcoratger, line 194: Here to avoid transforming again `addr_bound` to Felt, wouldn't it be better to create a variable at the start of the function containing

`(constants.get(ADDR_BOUND) orelse return HintError.MissingConstant)`

which is already a Felt and that you transform on the one hand into u256 for the purposes of the previous comparison then which you reuse here to compare with `addr`

src/hint_processor/math_utils.zig
```diff
@@ -0,0 +1,685 @@
...
+        try hint_utils.insertValueFromVarName(
+            allocator,
+            "y",
+            MaybeRelocatable.fromFelt((try x.div(Felt252.three()))
+                .sqrt() orelse Felt252.zero()),
+            vm,
+            ids_data,
+            ap_tracking,
+        );
+    }
```
tcoratger, line 285: Here as the only thing that changes is the `MaybeRelocatable` argument, can we create a variable for that and just call the `insertValueFromVarName` method outside of the if statement?

tcoratger (approved):

tcoratger (approved):

PR #508, opened 2024-04-19
add test filter for integration test

### LINKED ISSUE #454: Integration: Integration runner should take command line argument to run one or more tests
As we move to increase the surface area of integration tests, we want to be able to run a subset of all integration tests repeatably to aide in debugging iteration.

A possible solution is to have a command line argument that takes comma separated regex and only runs tests that have filenames that match the regex.

TropicalDog17:
Closes #454 

Example: ```make run-integration-test-filter FILTER="keccak"```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,10 @@ run-integration-test:
 	@zig build integration_test
 	./zig-out/bin/integration_test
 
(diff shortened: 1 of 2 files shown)


lana-shanghai (approved):
lgtm

PR #518, opened 2024-06-17
Fix field arithmetic benchmark cairo related bug + perf

StringNick:
- Field arithmetic benchmark after update on last version from cairo-vm(rust), panic an error with DiffAssertValues, after investigation founded that error was related with wrong getFromConstant logic.
- Improved perfomance of split64/128 functions.
- Added format for MaybeRelocatable type
- Fixed compile of field_arithmetic, added additional path to dependencies

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ docsite/
 cairo-vm-env
 cairo-vm
 tmp/
+profile.json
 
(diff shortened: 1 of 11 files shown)


lana-shanghai (approved):
lgtm

